PR #69, opened 2016-08-26
Updated Project and Subsidy data with APSH and DCHA info

katyaabaz:
Compared data between APSH, DCHA, Project, and Subsidy data. Updated
Project and Subsidy files with all DCHA suggestions. Updated Project and
Subsidy files with APSH data not conflicting with administrative data
(only if informal source). Noted remaining differences in program
commetns. Will post remaining differences (projects not in catalog) in
note on Box to be updated over time.


katyaabaz:
@NeighborhoodInfoDC/review

Issue #28, opened 2015-10-13
Figure out how to add new projects to the Catalog


ptatian:
There is currently no procedure for adding a new project to the Catalog, which would involve assigning it a unique ID and adding records to the appropriate related data sets.

ptatian:
Notes: https://urbanorg.app.box.com/notes/75277111378

ptatian:
Hi @mcohenui, I finally have something that we can use to try adding new projects. I've placed two files with new tax credit developments on L:

```
L:\Libraries\PresCat\Raw\Buildings_for_geocoding_2017-05-25_main.csv
L:\Libraries\PresCat\Raw\Buildings_for_geocoding_2017-05-25_subsidy.csv
```
I followed the templates you sent me, but did make one change. I believe you were assuming that people would run the addresses through the MAR geocoder before inserting them into the template worksheets, right? If so, it seems you would also need the image and Google Streetview URLs to be included, so **I added those two columns to the _main template**. If I've misunderstood something about the process, I can take them out. 

So, please test these out and see how it works! I'd like to review before we finalize anything. 

FYI, I'm almost ready to push out the tax credit project updates using the DHCD FOIA data (issue #25). I'll want to finalize those updates before we finalize adding the new projects. I'll be sending a separate note about that shortly. 

Thanks, Mychal. Please let me know if you have any questions.

mcohenui:
That is great news @ptatian. I will definitely carve out some time next week to start testing. Thanks for the update.

ptatian:
Hi Mychal, Just FYI, I will need you to add the new LIHTC projects to the Catalog before I can finish issue #22. Thanks.

ptatian:
@mcohenui As I mentioned, I made some modifications to the Project data set by adding three new variables:

- **proj_ayb** - Project year built (original)
- **proj_eyb** - Project year built (improvements)
- **proj_owner_type** - Project owner type (majority of parcels)

These variables will need to be filled in when adding new projects to the Catalog. Below is code that you can use to create these variables. This code assumes you are starting with complete Parcel data for one or more projects. In the example, I'm using the PresCat.Parcel file but it should also work if you substitute that with an updated parcel file for all the projects or just parcels for the new projects that you are adding. 

Similarly, at the end we merge the new variables back with the Project data. I'm using the PresCat.Project data set but you can substitute an updated temporary one. 

The program running this code needs to have access to both the PresCat and RealProp libraries. 

[...]

PR #75, opened 2016-09-15
Enhancement 28

mcohenui:
@ptatian This is the set of new programs/updates to existing programs meant to allow new projects to be added to the project data set.

New_Project_Geocode
Add_Subsidy
Create_real_property
Update_Parcel
Add_New_Project

Will email the relevant excel sheets.


ptatian:
Thanks, Mychal. I have received your request and will review. -Peter

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+
+**compare new datasets with old geocode datasets**;
+
+proc compare base=prescat.project_geocode compare=work.project listall maxprint=(40,32000);
+  id nlihc_id;
+run;
+
+proc compare base=prescat.building_geocode compare=work.building listall maxprint=(40,32000);
+  id Bldg_address_id;
+run;
```
ptatian: Good to use Proc Compare to check things. :+1:  But what are we looking for exactly? Will this be part of the permanent process, or just a checking things for now? If permanent, what should the person running the program be looking for?

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+
+** merge new geocode files onto existing geocode files**;
+
+data prescat.Project_geocode;
```
ptatian: Here we don't want to replace directly the existing files until it is the final batch run. So, you should save these as temporary (work) data sets until we are ready to finalize.

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+	set Project Project_geocode_a;
+	run;
+
+data prescat.Building_geocode;
```
ptatian: Same as comment above.

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+  where Ward2012 = '';
+  id nlihc_id;
+  title2 "&outlib..Building_geocode / Missing Ward";
+run;
+  
+proc print data=&outlib..Building_geocode noobs;
+  id nlihc_id;
+  var ssl Ward2012 Bldg_address_id bldg_addre;
+  title2;
+run;
```
ptatian: We can delete the four Proc Prints here. They are no longer necessary.
Instead, I would insert a couple of Proc Compares to check for changes in the new Project_geocode and Building_geocode data sets. 

> proc compare base=prescat.project_geocode compare=work.project_geocode listall maxprint=(40,32000);
>   id nlihc_id proj_name;
> run;
> 
> proc compare base=prescat.building_geocode compare=work.building_geocode listall maxprint=(40,32000);
>   id nlihc_id proj_name Bldg_address_id;
> run;

I added the Proj_name to the ID statements so that it is easier to identify the new projects in the output. Since Proj_name is the same for all obs with the same Nlihc_id, it doesn't affect the sort order.

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+data prescat.Project_geocode;
+	set Project Project_geocode_a;
+	run;
+
+data prescat.Building_geocode;
+	set Building Building_geocode_a;
+	run;
+
+%File_info( data=&outlib..Project_geocode, freqvars=Ward2012 Bldg_count )
+%File_info( data=&outlib..Building_geocode, freqvars=Ward2012 )
```
ptatian: Move these %File_info() to the %Register_metadata() macro, as we will need to run them on the permanent versions of the data sets for the final batch submit, but those files don't exist yet here.

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+  var ssl Ward2012 Bldg_address_id bldg_addre;
+  title2;
+run;
+
+** Macro Register_metadata - Start Definition **/
+
+%macro Register_metadata( revisions= );
+
+  %if &_remote_batch_submit %then %do;
+
```
ptatian: For finishing, as you've done we want to only register with metadata when it's the final batch submit. Before that, here's where we want to copy the temporary versions of Project_geocode and Building_geocode to replace the permanent data sets on L:. I would do that with a Proc Sort, which ensures that the final data sets are sorted properly. 

> proc sort 
>    data=Project_geocode 
>    out=PresCat.Project_geocode (label="Preservation Catalog, Project-level geocoding info");
> by nlihc_id;
> run;
> 
> proc sort 
>    data=Building_geocode 
>    out=PresCat.Project_geocode (label="Preservation Catalog, Building-level geocoding info");
> by nlihc_id Bldg_address_id;
> run;

Note that I've added the data set labels here, which are needed for the metadata system. 

Then let's do the %File_info() for each file. 

> %File_info( data=PresCat.Project_geocode, freqvars=Ward2012 Bldg_count )
> %File_info( data=PresCat.Building_geocode, freqvars=Ward2012 )

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+      creator_process=DC_info_geocode.sas,
+      restrictions=None,
+      revisions=%str(&revisions)
+    )
+    
+  %end;  
+  %else %do;
+  
+    %warn_mput( msg=Not final batch submit. Files will not be registered with metadata system. )
+
```
ptatian: Add the %File_info() for the temporary data sets here.

> %File_info( data=Project_geocode, freqvars=Ward2012 Bldg_count )
> %File_info( data=Building_geocode, freqvars=Ward2012 )

Prog/New_Project_Geocode.sas
```diff
@@ -0,0 +1,565 @@
...
+**************************************************************************/
+
+%include "L:\SAS\Inc\StdLocal.sas";
+
+** Define libraries **;
+%DCData_lib( PresCat )
+
+
+%let MAX_PROJ_ADDRE = 3;   /** Maximum number of addresses to include in Proj_addre field in PresCat.Project_geo **/
+%let outlib = %mif_select( &_remote_batch_submit, PresCat, WORK );
```
ptatian: I don't think we need the outlib macro var. We'll handle through a full macro instead.

mcohenui:
@ptatian I made the changes per your review, and committed the changes to branch #28.

ptatian:
Thanks. I'll review and get back to you.

ptatian:
Sorry, I didn't mean to approve all of the changes yet! (And not sure how to undo the approval.) Please stand by until I give further guidance.

ptatian:
Mychal, did you sync your changes with the remote repository? I don't see any commits since 13 days ago. Also, would you please place the latest versions of the Buildings_for_geocoding_2016-08-01_*.csv files in L:\Libraries\PresCat\Raw. Thanks!

ptatian:
Mychal, I don't seem able to start a new review, so I'll pass along my comments this way.

I made a few tweaks to the New_Project_Geocode.sas program, so **you should pull down (sync) my changes**.  One thing to note is that I moved all of the raw input files to L:, rather than D:. (That's preferred practice because otherwise people may forget to copy the files to L: later.) Anyway, I think that part is in good shape now, so we can move on to **Add_subsidy.sas**.

Good logic with what you've done so far, ie, using the address ID as a way of connecting the new subsidy and project records. But I think we shouldn't use the proj_address_id in Project or Project_geocode for this. That is only the "first" address ID, which is a bit arbitrary, so if the person doesn't pick that one the linking won't work. Rather, we should use the updated Building_geocode so that way the user can pick any address ID for the project and it will work.

Now, there's a little glitch with that idea because I added a check to the New_Project_Geocode.sas program for duplicate address id's across projects and it turns out there are several. This might be OK, because we could have a building that gets transferred from an older project to a new one. We do need to check on those later. But, for now, it will make SAS unhappy if you try to merge by address ID and there are multiples. Since the problem seems isolated to older projects, not new ones, I'd suggest doing a `proc sort nodupkey ...; by bldg_address_id;` on Building_geocode (to a new temporary data set), which will eliminate the duplicates. Then you can merge without a problem. (It's not necessarily the best solution, because the [nodupkey option](http://support.sas.com/documentation/cdl/en/proc/61895/HTML/default/viewer.htm#a000146878.htm) is a little arbitrary in how it picks which obs to keep. But, since I don't think it should be an issue with new projects it should be OK for now.)

A second challenge is, the way I'm now envisioning the process working, rather than running the programs in sequence we are going to combine them together so that it's just one big run. (I could be talked out of that if it seems not good for some reason.) Anyway, that means you can't assume there's an updated permanent Building_geocode data set yet. You have to use the temporary one that's created by New_Project_Geocode.sas. If you are doing your development in an interactive SAS session, that's not a problem, because you just need to run New_Project_Geocode.sas first to create the temporary data sets and then you can work from there. 

So, for the next phase, why don't you focus on getting the Add_Subsidy.sas program working using the approach above and then we can review and discuss further steps. 

Let me know if you have any questions or would like to go over it together.

mcohenui:
Peter, this sounds good. I don't immediately see any insurmountable issues with running the program in one large run. So I will work on the subsidy program with that set-up in mind. Thanks!

mcohenui:
@ptatian Committed changes to the add_subsidy program.

ptatian:
**Add_subsidy.sas** looks good. I pushed a couple of changes. Portfolio is actually derived from Program using this statement: `Portfolio = put( Program, $progtoportfolio. );`

You can go on to the **Project** file now. One part of updating **Project** is bringing in some summary data that are derived from **Subsidy**. You do this by executing this macro call in open code on the newly updated **Subsidy** data set:

`%Create_project_subsidy_update( data=Subsidy )`

This creates a new temporary data set called **Project_subsidy_update** that has updates of selected **Project** data set vars. You can then apply this file to the other updated **Project** variables, such as Proj_name, Status, etc. In my example below, these other vars have previously been updated in the **Project_update** data set.

```
data Project;

    update 
      Project_update (in=in_Project)
      Project_subsidy_update
        (keep=nlihc_id Subsidized Proj_Units_Assist_Min Subsidy_Start_First Subsidy_End_First 
      Proj_Units_Assist_Max Subsidy_Start_Last Subsidy_End_Last);
    by Nlihc_id;
    
    if in_Project;
    
run;
```

Note that the data set **Project_update** needs to include the full set of projects. 

Let me know if you have any questions.

mcohenui:
@ptatian Updated the commit with the revised programs for adding new projects:

New_Project_Geocode
Add_Subsidy
Update_Parcel
Add_New_Project

ptatian:
@mcohenui Sorry for the delay in getting to this. I'm taking a look now at the latest changes. Would you please walk me through the basic process flow? What steps do I follow and which programs do I submit, in what order, to add new projects to the Catalog? Thanks!

mcohenui:
@ptatian Created a very rough sketch of the process flow and the program order on box: https://urbanorg.box.com/s/5o6ykm5q0cb9fcy9td6esjqqfk5ev3ql.

PR #146, opened 2018-03-07
February Update

npstrayer:
@mcohenui Needs review

### DIFF
--- a/Prog/Updates/Update_MFIS_2018_02.sas
+++ b/Prog/Updates/Update_MFIS_2018_02.sas
@@ -1,0 +1,25 @@
+/**************************************************************************
+ Program:  Update_MFIS_2018_02.sas
+ Library:  PresCat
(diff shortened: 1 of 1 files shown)


mcohenui:
@npstrayer Looks good to merge.
@ptatian I notice the change to Belmont Crossing's ownership shows up in the output in SAS, but does not generate in the PDF printout "Update_MFIS_2018_02_project". Something to note.

ptatian:
Thanks, @mcohenui. Do you know if there was a record for Belmont Crossing in the HUD Sec8 file? The update report PDF shows records that got changed, but it might not include records that just have an exception without an actual update from the HUD source. (I'm speculating a bit because I would have to look at the code to tell if that's really how it works.)

mcohenui:
@ptatian No record in the Sec 8 file, just the MFIS. Makes sense then if the update report only shows changes from updates to the HUD files.

PR #150, opened 2018-03-30
Adding missing building and parcels

npstrayer:
@rpitingolo Running into an error when I try to use the MAR_geocode macro 

Thanks


npstrayer:
Hi Peter, 

I've placed a copy of the MAR addresses on the L drive here:

L:\Libraries\PresCat\Raw\MAR_addresses.csv

Thanks,
Noah

PR #180, opened 2018-11-07
Update Sec8 for Catalog

Woliver01:
@ptatian please review.


ptatian:
Thanks, @Woliver01. I will review. 

You can just put the review request message in the pull request (that is, here). You don't need to include it in the commit message as well.

PR #197, opened 2019-04-03
Update #196

Woliver01:
@mcohenui Please review.


mcohenui:
@Woliver01 There was an error with the date formats for the REAC file. I went in and made a fix to the formatting in the raw file in the HUD repository. That is something we should watch out for in the next new REAC file we get. But for now this should be good to merge.

PR #223, opened 2019-11-26
Adding LIHTC projects at end of year 15 expiring subsidy

Woliver01:
@mcohenui I believe I got the code to produce the LIHTC 15yr data but I'm not sure that it's also producing the 30yr expiration dates as well. I may be wrong, but I think I need to create a new data statement and then merge those two together. I tried playing around with the code and couldn't quite get it to produce what I thought it should but I may be wrong in needing to do this step. Please review.

### DIFF
--- a/Prog/Reports/Expiration_rpt.sas
+++ b/Prog/Reports/Expiration_rpt.sas
@@ -63,8 +63,9 @@ proc report data=Expiration_rpt nowd
       style(header)=[fontsize=2 font_weight=bold textalign=left]
       style(column)=[fontsize=2 textalign=left];
   by Category_code;
-  column rpt_id poa_end program;
+  column rpt_id poa_end compl_end program;
   define rpt_id / "Project" display;
+  define compl_end / display;
   define poa_end / display;
   define program / display;
   label category_code = 'Category';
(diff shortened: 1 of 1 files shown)


mcohenui:
@Woliver01 Hey Will, I am working remotely so can't access the datasets to run it on my laptop. But looking at the code, you are basically right. You need one data set, with both sets of data. You could do this by adding the 15 year "where" statement into the first data step. The way that it is coded now, the second data step with the 15 year data just overwrites the 30 year data.

Woliver01:
@mcohenui Just submitted new code, please review. Also let me know if you need to see the spreadsheet, I can email it to you.

mcohenui:
@Woliver01 I went ahead and updated the merge statement. It needed to have both the poa_end statement and the compl_end statement. The first part should be good to go. You now need to edit the excel output so that it identifies the 15 year properties.

Woliver01:
@mcohenui okay. Should I merge this branch before continuing so that it brings those changes to the program?

mcohenui:
@Woliver01 If you go to your github desktop, and go to this branch, you should be able to pull my commit onto your local drive.

Woliver01:
@mcohenui I updated the code for the output table. Please review.

mcohenui:
@Woliver01 Looks good Will. I only changed the program to group based on program type, so it will be easier to identify LIHTC buildings.

PR #231, opened 2020-01-13
Merge VCU/CNHED LEC database with PresCat

ptatian:


ptatian:
LECs will be identified in the Catalog by a subsidy file record with the program label, “Limited equity cooperative.” For the 32 existing projects, this will be an additional subsidy record to those already on the project (such as for Sec 8, LIHTC, etc.). For the new projects, this will be the only subsidy record associated with the project. 

Other fields in the subsidy record will be filled in from data in the worksheet as follows. Items in quotes (“”) are constant values that will be assigned to all records.
- Poa_start (start of affordability) &larr; Year Co-op Formed. One project has an exact date so we will use that, but for others only the year is provided so the date will be set to Jan 1 of that year. If the year is missing entirely (N=2), the Poa_start field will be empty.
- Units_assist (assisted units) &larr; # of Units. If # of Units is missing in the spreadsheet (N=1), will fill in using existing unit count info from Catalog.
- Subsidy_info_source &larr; “VCU-CNHED/Limited equity cooperative database”
- subsidy_info_source_date &larr; “1/10/2020”
- Poa_end (end of affordability). Will be empty since no affordability expiration date for LECs. 

Fields in the project record will be filled in as follows:
- proj_name (project name) &larr; Cooperative
- hud_mgr_name (property manager) &larr; Management
- hud_own_name (property owner) &larr; [Owner name taken from DC real property database]
- hud_own_type (property owner type) &larr; “Non-profit”
- Status &larr; “Active”
- Subsidized &larr; “Yes”
- Category_code (Pres Cat category). For existing projects, current category will be kept. For new projects, category will be set to 5 (Other Subsidized Property).
- Geographic identifiers (ward, ANC, etc.) will be filled in using standard Catalog approach based on geocoded street address.

PR #238, opened 2020-03-06
HUD data for March PresCat Update

Woliver01:
There was an invalid input request in the REAC score log. @mcohenui please review.


mcohenui:
@Woliver01 The REAC dates have the wrong formatting. They should be formatted mm/dd/yyyy, but are instead formatted as dd-mmm-yy. To fix it, you can either create a issue to reformat the reac files in the HUD repository or you can change the formatting on the raw file (saving it as a copy) and rerun the HUD reac update program.

Since it seems as though HUD is being inconsistent with date formatting, you should flag for the next update.

PR #247, opened 2020-10-05
Adding IZ units to Catalog

mcohenui:
@ptatian  I had a few questions about adding the IZ units to the catalog. The current dev program you had for adding the coops, has a number of checks against the catalog. My understanding is that there should not be any IZ units that are already in the catalog, however, the check using SSLs found a number of matches. What do you think might be causing that? Am I correct in understanding that there should not be any matches between the new IZ data and existing projects in the catalog? I could also be missing something given my relatively rudimentary understanding of the SQL calls that are in the program.

### DIFF
--- a/Prog/Dev/242_Review_IZ_db.sas
+++ b/Prog/Dev/242_Review_IZ_db.sas
@@ -59,6 +59,13 @@ data IZ_db;
     otherwise Address_ref = Address;
   end;
   
+  **Create AMI categories**;
+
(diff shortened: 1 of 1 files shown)


ptatian:
Hi @mcohenui. I would have to look more closely but you are right it is unexpected that there should be much overlap, given most of the Catalog consists of older projects. It's possible that some newer subsidized developments include IZ units, if they are mixed income. It also may be that the IZ units are in a new development that is located where an assisted project used to be. 

Can you email me a few examples so I can take a look?

PR #249, opened 2020-10-09
Update PresCat with Sec8 data

Woliver01:
@mcohenui There seems to be an error with this line of code in the macro, and I'm not sure how to trouble shoot: 
%DCData_lib( General, env=windows, macdef=dcautos uiautos sasautos ).

Please review.

### DIFF
--- a/Prog/Updates/Update_MFIS_2020_09.sas
+++ b/Prog/Updates/Update_MFIS_2020_09.sas
@@ -20,7 +20,7 @@
 %DCData_lib( PresCat )
 %DCData_lib( HUD )
 
-
+options nominoperator;
 %Update_MFIS( Update_file=MFIS_2020_09 )
 
 
--- a/Prog/Updates/Update_MFIS_2020_09.sas
+++ b/Prog/Updates/Update_MFIS_2020_09.sas
@@ -1,0 +1,26 @@
+/**************************************************************************
+ Program:  Update_MFIS_2020_09.sas
+ Library:  PresCat
+ Project:  Urban-Greater DC
+ Author:   W.Oliver
+ Created:  10/9/2020
+ Version:  SAS 9.4
+ Environment:  Local Windows session (desktop)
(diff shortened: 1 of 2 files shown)


mcohenui:
@Woliver01 what is the error you are getting? I am not getting that error on my end, but I am seeing a few errors on the update_history_recs macro.

Woliver01:
@mcohenui So I closed out the program and re-ran it, so that particular error no longer showed up but now I'm getting a ton of errors. The first of which being these two:
ERROR: Operand missing for IN operator in argument to %EVAL function.
ERROR: The macro SUPER_TRANSPOSE will stop executing.

mcohenui:
@Woliver01 Yea, those were the errors I was getting. And I think they are related to the issue with the Access file. The variables in the excel were not in the same order as the variables from the access file. So if you look at the sec8mf_2020_09_dc file some of the variables have the wrong data in them.

The easiest solution would be to create a copy of the excel (saving the original raw copy), edit the file so that the variables align, test the program to ensure the data looks correct, and then re-batch submit the Sec8mf_2020_09 program. It should not require another github issue, since you shouldn't actually need to edit the program itself. Let me know if you want to talk through it.

Woliver01:
@mcohenui Yeah, could we hop on a call quickly?

Woliver01:
@mcohenui Hey Mychal, I got the same errors for the MFIS update. Please review.

mcohenui:
@Woliver01 Yea I am getting the same issue. @ptatian We are getting a number of odd errors with the PresCat update. I am unable to identify what could be causing them. Will could you send a copy of the log via email?

rpitingolo:
@mcohenui @Woliver01 I can take a look since Peter is out at the moment. What is the main program that you run for this update?

rpitingolo:
@Woliver01 @mcohenui please pull down my commit and try running it again.

PR #252, opened 2020-11-09
Create Update_Sec8mf_2020_10.sas

Woliver01:
@mcohenui Hey Mychal, I believe there's an error with the macro for the Sec_8 program but I wanted you to take a look to make sure that's the correct assessment. And if so, which macro program is the one that needs to be edited?

### DIFF
--- a/Prog/Updates/Templates/Update_Sec8mf_yyyy_mm.sas
+++ b/Prog/Updates/Templates/Update_Sec8mf_yyyy_mm.sas
@@ -14,12 +14,12 @@
  Modifications:
 **************************************************************************/
 
-%include "L:\SAS\Inc\StdLocal.sas";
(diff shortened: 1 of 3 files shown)


mcohenui:
@Woliver01 looks like it is the issue with the 'IN' operator that happened last month with the MFIS data. Try adding "options nominoperator;" to update program before the macro call and see if that fixes it. Shouldn't have to alter the macros at all. Also you should update the Sec8 and MFIS templates so that they include this addition for future updates.

Woliver01:
@mcohenui Do you mean add this before the %include statement?

mcohenui:
@Woliver01 before the " %Update_Sec8" statement

Woliver01:
@mcohenui That worked and I added the lines of code to the template. Please review.

mcohenui:
@Woliver01 good to merge.

PR #259, opened 2021-02-03
Create Update_Sec8mf_2021_02.sas

Woliver01:
@mcohenui There was an error with the program registering the new data as a separate month/update. Please review.

### DIFF
--- a/Prog/Updates/Update_MFIS_2021_02.sas
+++ b/Prog/Updates/Update_MFIS_2021_02.sas
@@ -1,0 +1,27 @@
+/**************************************************************************
+ Program:  Update_MFIS_2021_02.sas
+ Library:  PresCat
+ Project:  Urban-Greater DC
+ Author:   W. Oliver
+ Created:  02/03/2021
+ Version:  SAS 9.4
+ Environment:  Local Windows session (desktop)
+ GitHub issue:  
+ 
+ Description:  Update Preservation Catalog with latest 
+ HUD MFIS update file.
+
+ Modifications:
+**************************************************************************/
+
+%include "\\sas1\DCdata\SAS\Inc\StdLocal.sas";
+
+
+** Define libraries **;
+%DCData_lib( PresCat )
+%DCData_lib( HUD )
+
+
+%Update_MFIS( Update_file=MFIS_2021_01 )
+
+
--- a/Prog/Updates/Update_REAC_2021_02.sas
+++ b/Prog/Updates/Update_REAC_2021_02.sas
@@ -1,0 +1,26 @@
+/**************************************************************************
+ Program:  Update_MFIS_2021_02.sas
+ Library:  PresCat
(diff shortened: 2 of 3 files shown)


mcohenui:
@Woliver01 It looks like you are calling an update file named "Sec8mf_2021_02", but the file updates in the HUD repository are labeled Sec8mf_2021_02.sas. So the program is actually not finding the update file, and then indicating that the update is not newer that the last update. If you fix the update file call, it should run through.

Woliver01:
@mcohenui Do you mean to change the call in update_sec8mf_yyyy_mm to January instead of February?

mcohenui:
No the actual content of the macro: %Update_Sec8mf( Update_file=Sec8mf_2021_02 )

Woliver01:
@mcohenui Do you mean to change %Update_Sec8mf( Update_file=**Sec8mf_2021_02**) to %Update_Sec8mf( Update_file=**Sec8mf_2021_01**) in the update_sec8_yyyy_mm program? Or do you mean the sec8mf macro? Because there are 5 update_sec8mf programs and I could use some clarification:
update_sec8mf
Update_sec8mf_finish
update_sec8_init
update_sec8mf_project
update_sec8mf_subsidy

Woliver01:
@mcohenui I meant 5 update_sec8mf macros.

mcohenui:
@Woliver01 The former. I went ahead and committed the edit, just for clarity. Basically the way the program is written, the macro call (designated by the %) is calling a file that does not exist.

Woliver01:
@mcohenui Okay thanks. But I guess I'm confused because I did name the program sec8mf_2021_**02** because we already had a program name sec8mf_2021_**01** from the previous update, and we wanted to differentiate this current update from the previous one. Was the reason for having to change the macro you just edited have to do with the update date (on the HUD website and thus entered into the sec8mf_yyyy_mm program) being 01/28/2021 and not actually in February (as the name of the sec8mf_2021_**02** program would suggest) ?

Woliver01:
@mcohenui Please review.

mcohenui:
@Woliver01 This might be easier to explain through a call, but I will take a crack at it here and if you have additional questions feel free to call me.

So the naming of the program itself is different then the naming of the output. So for instance, even though the HUD update program is named secmf_2021_02, the resulting data file is named based on the file update date (ie 1.28.21), because that is what is entered into the macro call in the program. So if you look into the "Data" folder in the HUD repository, you will see that all of this month's updates are labeled _2021_01.

That means that when you moved to the prescat update and tried to call the _2021_02 files, the system did not find any. Even though  you saved those programs as _2021_02.

In short, the name of the program describes the month of the update, but the name of the resulting data file takes on the month of the data.

mcohenui:
@Woliver01 also good to merge

Woliver01:
@mcohenui okay thanks! That makes sense.

PR #275, opened 2022-04-11
Sec8MF Update

dharvey17:
@ptatian

### DIFF
--- a/Prog/Updates/Update_Sec8mf_2022_01.sas
+++ b/Prog/Updates/Update_Sec8mf_2022_01.sas
@@ -1,0 +1,25 @@
+/**************************************************************************
+ Program:  Update_Sec8mf_yyyy_mm.sas
+ Library:  PresCat
+ Project:  Urban-Greater DC
+ Author:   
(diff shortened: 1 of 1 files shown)


ptatian:
@dharvey17 Are you ready for a review on this?

dharvey17:
Yes! Sorry about that - did not make it explicit

ptatian:
Yes, you should tag me. I'll look at it.

Prog/Updates/Update_REAC_2022_02 (1).sas
```diff
@@ -0,0 +1,25 @@
+/**************************************************************************
```
ptatian, line 1: Not sure how it happened, but would you please rename the program file without the " (1)"?

ptatian (changes requested):
@dharvey17 This looks good! Would ask you to rename the REAC program so that it doesn't have the "(1)" in the file name. But then you can merge and do the final batch submit on SAS1. Let me know if you have questions. Thanks!

Issue #290, opened 2022-07-08
Add new Census geos, neighborhood clusters, and wards to PresCat.Building_geocode


ptatian:
This task is to add the 2020 census geographies (tracts, block groups, blocks), 2017 neighborhood clusters, and 2022 wards to the Preservation Catalog data sets and code. The first step is to add the new geography variables to the **PresCat.Building_geocode** data set, which has the most detailed set of locations and is the source for updating geographies in other data sets. 

Since the new geographies have been added to the MAR addresses, and since all of the records in PresCat.Building_geocode have a MAR address ID (MARID), the best way to do this is to merge the new geographies from the MAR data using ADDRESS_ID.

ptatian:
Since we are updating a permanent data set, you will need to use the %Finalize_data_set() macro at the end of your program. This DCData macro handles the special processing we need for the metadata and also won't replace the existing data set until we do the final batch submit on SAS1 after the code has been reviewed and merged. 

```
  %Finalize_data_set( 
    /** Finalize data set parameters **/
    data=Building_geocode,
    out=Building_geocode,
    outlib=PresCat,
    label="Preservation Catalog, Building-level geocoding info",
    sortby=Nlihc_id Bldg_addre,
    /** Metadata parameters **/
    revisions=%str(Add new Census geos, neighborhood clusters, and wards.),
    /** File info parameters **/
    printobs=10 
  )
```

This macro call assumes you have created a temporary data set called Building_geocode (i.e., Work.Building_geocode) that has the new variables added to it. When you are working in interactive SAS (in the Windows app) or doing a local batch submit the temporary data set will not be saved. But on the final submit on SAS1 the temporary data set will replace the permanent one currently on SAS1. 

Let me know if you have any questions. Thanks!

PR #307, opened 2022-07-19
Update #290

eburton9:
@ptatian could you review? seeming like things are working from the log, but in the output viewer I don't see any of the new geographies..

### DIFF
--- a/Prog/AddNew/Add_new_geographies_2020.sas
+++ b/Prog/AddNew/Add_new_geographies_2020.sas
@@ -7,9 +7,8 @@
  Version:  SAS 9.4
  Environment:  Local Windows session (desktop)
  
- Description:  Add new geographies to Preservation Catalog. 
+ Description:  Add 2020 census geographies, 2017 nhbd clusters, 2022 wards to Preservation Catalog. 
  
-
  Modifications:
 **************************************************************************/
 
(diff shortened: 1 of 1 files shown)


ptatian:
Thanks, @eburton9  Taking a look.

Prog/AddNew/Add_new_geographies_2020.sas
```diff
@@ -0,0 +1,48 @@
...
+	merge Building_geocode (in=a) 
+		address_points_2022_07 (in=b);
+	by bldg_address_id;
+
+	if a then output; 
+run;
+
+  %Finalize_data_set( 
+    /** Finalize data set parameters **/
+    data=Building_geocode,
```
ptatian, line 39: @eburton9 The data= parameter is the data set that will eventually be finalized. You are using the Building_geocode which is the original file. Your updated file is called left_join, so that's the name you should put here.

ptatian (changes requested):
@eburton9 A couple of comments--one that will address why you aren't seeing the new variables at the end. Please let me know when you've committed those changes and I can check again. Thanks!

eburton9:
@ptatian thank you! made these changes, the output looks correct and no errors.

ptatian (approved):
@eburton9 Looks good! You can go ahead and merge this pull request. Then sync on SAS1 and do a final batch submit there. Thanks!

PR #312, opened 2022-07-21
Update Project_category_view.sas

eburton9:
@ptatian ready for review


ptatian (approved):
@eburton9 This looks good! One thing that I forgot earlier was to include a variable with the neighborhood cluster names, like we had in 2000. That can be generated in the SQL code, so I added that. This is OK to merge and batch submit on SAS1. Thanks!

PR #313, opened 2022-07-22
update geos in macros

eburton9:
@ptatian ready for review


ptatian (approved):
Thanks, @eburton9. This is good to go.

PR #316, opened 2022-07-29
Cleaning TOPA pipeline database

eburton9:
@ptatian This is as far as I got, apologies I was hoping to do more. I'm in the midst of changing variables from char to num, but have some commented out because I was trying to figure out why the drop function wasn't working. It still isn't dropping the variables though. At the bottom of the code the rest of the variables that need to be converted to num are there as well. Thanks!


dharvey17:
@ptatian ready for review - output looks good on my end!

ptatian (approved):
Thanks, @dharvey17 . This is ok to merge and batch submit on SAS1.

ptatian:
I cleaned up a few things, such as deleting the obsolete code that was commented out and dropping an extra variable that was being created by Proc Import.

dharvey17:
\\sas1\DCDATA\Libraries\PresCat\Prog\TOPA\DC_Pipeline_2022_07.log

PR #322, opened 2022-10-04
address parse and mar geocode

eburton9:
@ptatian two issues: 1) the keepin= parameter for the %Rcasd_address_parse( ) macro isn't keeping the variables you specified for the TOPA addresses data set, there isn't an error in the log; 2) a few addresses from CNHED's sheet aren't working in the address parse marco (noticed after trying to geocode), I went through the output and noticed it was when there were different streets among the addresses (see ID 73 in the google sheet, highlighted in red). 

Thanks!


ptatian:
@eburton9 Thanks for flagging. You are right there was no "error." I forgot that _keepin=_ only works on the input data, not the output data. Because other code uses this macro I created a separate _keepout=_ parameter that can be used to add the two variables to the output data set. 

I updated your program with the change needed. I also used an _(obs=)_ option on the first proc print so that it doesn't try to print the entire data set anymore (you can adjust the number of obs). I replaced the other two proc prints with %File_info() macro calls, also limiting the number of obs (_printobs=_). The advantage of %File_info(), which is an Urban macro library macro, is that it will label the output with the file name and also produces a proc contents, which can be helpful to review for things like variable types and lengths, and proc means, which can be used to look at numeric value ranges. You can suppress the proc contents by adding _contents=n_ to the macro call and suppress the proc means by adding _stats=_ (nothing following the = sign). 

You should be OK to proceed from here. Let me know if you have more questions. Thanks!

ptatian:
@eburton9 Sorry, should have mentioned that the changes I made to %Rcasd_address_parse( ) should work for you automatically, unless you have a local version of the DHCD repo on your C: or D: drive. If you do have a local repo, then you should pull the changes on the main branch to get the updated macro code.

ptatian:
@eburton9 Sorry again. I forgot the second issue about the addresses. This is not a surprise as the addresses in the RCASD lists are often very irregular. The macro tries to deal with most of that irregularity, but there are things it doesn't know how to deal with. 

For the moment you can ignore the addresses that don't work. I'll take a closer look at those and come up with some next steps to address it.

eburton9:
@ptatian tried to do the last step on the instructions.. 1) not sure which variables should be kept from Real.Prop.Sales_master dataset when matching to TOPA? 2) Is there a way to do some sort of filtering when using proc sql to match datasets to only keep the sales records that have dates after "CASD Report week ending date" and "Offer of Sale date? (should that be compared to saledate in Real.Prop.Sales_master and that date should come after both CASD and offer of sale date?)

ptatian:
@eburton9 1) You should keep the sale date, sale price, owner name, owner address, and ui_proptype variables. 2) Try adding a WHERE statement to the Proc SQL to select only those obs where the sale date is after **the later of** "CASD Report week ending date" and "Offer of Sale date?" With SAS date values, "later" can be expressed using the max() function.

ptatian:
@eburton9 Although, I meant to update you that I had a conversation last week with CNHED and what might be better is to find two sales: (a) the last sale before the earlier of "CASD Report week ending date" and "Offer of Sale date" and (b) the first sale after the later of "CASD Report week ending date" and "Offer of Sale date." To be honest, I'm not sure the best way to do that. One way might be to start with the list of all post sales (original task) and then separately find the earliest one. 
Why don't you proceed with the original task and then see if we can come up with ideas .

eburton9:
max part for which one is later in time and then after part comparing that date to sale date (if sale date is after max date than greater than sale date is greater than the max of csad and notice date date after casd rreport date and notice date

eburton9:
@ptatian ready for your review!

Prog/AddNew/TOPA_data_to_DC_data.sas
```diff
@@ -0,0 +1,155 @@
...
+run;
+
+proc print data=TOPA_database1 (obs=5); 
+run; 
+
+data TOPA_database;
+    set TOPA_database1;
+    CASD_date = input(CASD_Report_week_ending_date, MMDDYY10.);
+    format CASD_date MMDDYY10.;
+    drop CASD_Report_week_ending_date;
```
ptatian, line 73: I wouldn't drop the original vars as it can be helpful to see them if we need to check the date coding.

ptatian: It would also be helpful to generate a list of notices with missing dates. Would you please add a Proc Print to the program notices that are missing sale date and CASD report date that we can share back with CNHED? We don't want to  include all variables--just key ones that will help them identify those notices.

Prog/AddNew/TOPA_data_to_DC_data.sas
```diff
@@ -0,0 +1,155 @@
...
+  create table TOPA_realprop as   /** Name of output data set to be created **/
+  select
+    coalesce( TOPA_SSL.SSL, RealProp.SSL ) as SSL, /** Matching variables **/
+	/** obs where sale date is after the later of CASD data and offer of sale data **/
+    TOPA_SSL.ID, /** Other vars you want to keep from the two data sets, separated by commas **/
+	TOPA_SSL.CASD_date, TOPA_SSL.offer_sale_date,
+	TOPA_SSL.Anc2012, TOPA_SSL.cluster2017, TOPA_SSL.Geo2020, 
+	TOPA_SSL.GeoBg2020, TOPA_SSL.GeoBlk2020, TOPA_SSL.Psa2012,
+	TOPA_SSL.VoterPre2012, TOPA_SSL.Ward2022, 
+    RealProp.SSL, RealProp.SALEPRICE, RealProp.SALEDATE, RealProp.OWNERNAME, RealProp.ui_proptype,
```
ptatian, line 133: This seems to be working well! For the RealProp vars, let's change to using the following: SSL, saleprice, saleprice_prev, saledate, saledate_prev, ownername_full, ownername_full_prev, ui_proptype, address1, address2, address3, address1_prev, address2_prev, address3_prev.

Prog/AddNew/TOPA_data_to_DC_data.sas
```diff
@@ -0,0 +1,155 @@
...
+quit;
+
+%File_info( data=TOPA_SSL, printobs=5 )
+
+%File_info( data=RealProp.Sales_master, printobs=5 )
+
+** match property sales records in realprop.sales_master to TOPA addresses **;
+proc sql noprint;
+  create table TOPA_realprop as   /** Name of output data set to be created **/
+  select
```
ptatian, line 125: Add the `unique` keyword after select here. This will filter out duplicate records from the final data set.

ptatian:
@eburton9 This all is coming along well! I've added some comments for changes. Please let me know when you have made those.

eburton9:
@ptatian Made the above changes. I included dates where either "CASD Report week ending date" and "Offer of Sale date" were missing rather than both since only one entry is missing both.

ptatian:
Thanks, @eburton9. A couple of additional things in prep for tomorrow's meeting. I'd like to create a spreadsheet with the sales data that we can go through with them. We'll try using the SAS XP tagset to do this as I think it will let us format the output in an easier to use way.

First, you need to set up the XP tagset by batch submitting Prog\Add_ExcelXP_tagset.sas in the General library on SAS1. You only need to batch submit this program once, and the tagset will be set up. 

Then you can add code to your program to export the data as a spreadsheet. 

```
** Export 2007 TOPA/real property data **;

ods tagsets.excelxp   /** Open the excelxp destination **/
  file="&_dcdata_default_path\PresCat\Prog\AddNew\TOPA_data_to_DC_data_2007.xls"  /** This is where the output will go **/
  style=Normal    /** This is the ODS style that will be used in the workbook **/
  options( sheet_interval='bygroup' )   /** This creates a new worksheet for every BY group in the output **/
;

ods listing close;  /** Close the regular listing destination **/

proc print data=TOPA_realprop;  /** Create the output for the workbook **/
  where year( casd_date ) = 2007;  /** Only use 2007 data **/
  by id;   /** BY groups (worksheets) will be for each TOPA ID **/
run;

ods tagsets.excelxp close;  /** Close the excelxp destination **/
ods listing;   /** Reopen the listing destination **/
```

This should create a file, TOPA_data_to_DC_data_2007.xls, that acts like an Excel spreadsheet. It is not a real Excel file, however, so when you open it in Excel you will get a warning message that it is not in Excel format. However, if you just say OK then it will open in Excel. 

I limited to 2007 to make the workbook more manageable. We could also do a later year -- that might be good to see the contrast.

Let me know if you have questions about the above. Thanks!

ptatian:
On Prog\Add_ExcelXP_tagset.sas, you can batch submit this right on SAS1. You don't need to copy or download it anywhere.

eburton9:
@ptatian created the excel file, do you want it saved anywhere other than my local drive as of now?

ptatian:
Thanks, @eburton9. Can you please put it [on Box](https://urbanorg.app.box.com/folder/180956887756)?

ptatian:
@eburton9 Yes, the idea was to have each ID in a separate worksheet. I thought that might be easier for scanning.

ptatian:
@eburton9 Sorry, would you please drop the geography variables from the workbook output?

eburton9:
@ptatian done, i'll update the box file too

eburton9:
@ptatian the excel is working with saledate and id sorted but I was wrong this morning, it looks like the days range format you created isn't being read into the proc tabulate.

ptatian:
Hi @eburton9 . Oops! I made a mistake on the format statement. Can you see what it is? (It's a classic SAS error!)

eburton9:
@ptatian ah! I had even checked semi colons and commas but missed the period. it looks good to me now!

ptatian:
@eburton9 Would you please change the location of the Excelxp file to `"&_dcdata_default_path\PresCat\Prog\AddNew\TOPA_data_to_DC_data_2007.xls"`? Other users won't be able to access the path you have there now. 

Also, just a few small tweaks to the proc tabulate table statement (for appearance). 
```
  table 
    /** Rows **/
    n='Notices of sale filed'
    n=' ' * days_notice_to_sale='Days between notice and property sale' 
    ,
    /** Columns **/
    all='All years' offer_sale_date='By notice year'
  ;
```

eburton9:
@ptatian changed file path and edited proc tabulate statement

ptatian:
@eburton9 I fixed most of the address problems that we were having. Some of it was changes I pushed to the %Rcasd_address_parse() macro. Others were some manual edits I made to addresses in the data. 

If you don't have a local version of the DHCD repo on your C: or D: drive, then you should be good to go. If you do have local DHCD repo, then you will need to pull down the changes from the main branch. After that, if you rerun the program, you should not see as many address issues.

ptatian:
I'll follow up with next steps tomorrow.

ptatian:
Thanks, @eburton9 . Here are some thoughts on final steps. 

First, the table that we made isn't quite right. I should have thought about the unit of observation. We want the unit to be a notice, but right now it is the transactions. So, let's fix that. 

Instead of using Topa_realprop as the input data for the table, we should summarize that data by notice ID first. We can create a new summary data set with Proc Summary. 

```
proc summary data=Topa_realprop;
  by ID;
  id ward2022;
  var offer_sale_date days_notice_to_sale;
  output out=Topa_realprop_by_id min=;
run;
```

Then you would use **Topa_realprop_by_id** in the Proc Tabulate. Note that I'm using the **min=** stat in the output statement. That will save the minimum values of the variables listed in the **var** statement in the summary data set. For offer_sale_date, the values are the same for an ID value it's just keeping that date. For days_notice_to_sale, however, that will give us the number of days to the earliest sale after the notice. 

I've also added the ward2022 var to the **id** statement, which will copy the ward value to each record. That's because I'd like to use that in a second table (below).

ptatian:
For the table, let's also output it to Excelxp, because that will be nicer looking and easier to share. You should copy the ODS statements you used to create the workbook with the 2007 sales, and modify them for the table output. Like before, you want to put the first two ODS statements that open Excelxp and close listing _before_ the Proc Tabulate, and then the two statements that close Excelxp and open listing _after_. 

For the **ods tagsets.excelxp** statement, you will want to give the table output file a different name. Also, we want to change the **sheetinterval=** to **'proc'**, which will output separate sheets for each procedure...

Because we are going to add a second table, which will be in a second worksheet. For the second table we will replace the notice years (offer_sale_date) with the 2022 wards. To do this, just copy the current Proc Tabulate in its entirety and make a second version right after it. You can then modify the second one to create the new table with wards. Note that the ward2022 variable needs to be listed in the **class** statement as well. 

The default tab names that SAS generates aren't very helpful, but we can manually change the sheet names by inserting a statement like this _before_ the procedure creating the output:

`ods tagsets.excelxp options(sheet_name="By year");`

You would put the above before the first Proc Tabulate and then a second one, with an appropriate label, before the second Proc Tabulate. These are in addition to the **ods tagsets.excelxp** statements that open and close the workbook. 

(There's a lot you can do with the Excelxp destination. [Here are some additional examples](http://support.sas.com/rnd/base/ods/odsmarkup/excelxp_demo.html) to check out.)

ptatian:
There is one more change I'd like to ask for, which is to add the 2012 wards to the output data set. I suspect that the wards in the Google Sheet are 2012, so lets have both in the output data. To do this, you need to add a **keep_geo=** parameter to the %DC_mar_geocode() macro. This parameter is used to list all the geo variables that you want to have in the geocoded data set. You can just list the current geo variables that we are using in the Proc SQL and add Ward2012 to that list.

ptatian:
Sorry, one other small thing. In your **data TOPA_database;** data step, please add these statements immediately after the **set** statement.
```
format _all_ ;
informat _all_ ;
```
Proc import adds a lot of extraneous formats and informats to data sets, so I like to clean that up. These statements will delete them. (In a format/informat statement, if you don't put anything after a variable name or list of vars, that removes the formats/informats from the variable(s). **\_all\_** is a SAS keyword that stands for all variables in a data set.)

ptatian:
OK. Last thing is to finalize the data set. I see you already have that code in the program, so you can just uncomment it. The only change is specifying the **sortedby=**, which should be ID. 

But let's also keep the **TOPA_realprop** data set as well. Please add a second %Finalize_data_set() macro for that data set. It should be sorted by ID and saledate, as we did earlier.

BTW, you didn't need to comment out the %Finalize_data_set() code earlier. This code is designed so that the data set will be saved permanently only when you batch submit the program on SAS1. As long as you are working on your local version of the program on C: or D:, no permanent data sets will be created or replaced. 

Thanks, @eburton9 . Sorry for the lengthy instructions, but I think we should be able to close this pull request once these steps are done and I give it one last review. Please let me know if you have any questions.

eburton9:
Thanks, Peter! I appreciate all the SAS notes throughout the instructions (slowly learning!). Made the above updates and the two excel sheets and output datasets look good to me.

Prog/AddNew/TOPA_data_to_DC_data.sas
```diff
@@ -0,0 +1,287 @@
...
+  create table TOPA_realprop as   /** Name of output data set to be created **/
+  select unique 
+    coalesce( TOPA_SSL.SSL, RealProp.SSL ) as SSL, /** Matching variables **/
+	/** obs where sale date is after the later of CASD data and offer of sale data **/
+	saledate - offer_sale_date as days_notice_to_sale,
+	TOPA_SSL.ID, /** Other vars you want to keep from the two data sets, separated by commas **/
+	TOPA_SSL.CASD_date, TOPA_SSL.offer_sale_date,
+	TOPA_SSL.Anc2012, TOPA_SSL.cluster2017, TOPA_SSL.Geo2020, 
+	TOPA_SSL.GeoBg2020, TOPA_SSL.GeoBlk2020, TOPA_SSL.Psa2012,
+	TOPA_SSL.VoterPre2012, TOPA_SSL.Ward2022,
```
ptatian, line 140: Here as well.

ptatian (changes requested):
@eburton9 This looks good. I looked at the summary tables and pushed a couiple of quick changes so that the total number of notices would equal the total in the TOPA database. Some of them get dropped at different points, along the way. 

With that in mind, it would be helpful to save TOPA_addresses and TOPA_ssl as permanent data sets, too. When we go to the next stage, we may need all of them. Would you please make that change?

eburton9:
@ptatian Finalized TOPA.SSL and TOPA.addresses at the end and added Ward 2012 to the two proc sqls.

Issue #326, opened 2022-11-11
Preservation Catalog PresCat Dec 2022 update


ptatian:
_Note: If you have not previously done so, [clone the PresCat repo](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/adding-and-cloning-repositories/cloning-and-forking-repositories-from-github-desktop#cloning-a-repository) to **C:\DCData\Libraries**. Pull the latest changes from main branch before proceeding._

## Apply updates to Catalog

- [x] [Create new branch](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/making-changes-in-a-branch/managing-branches#creating-a-branch) **update-#**_n_ in GitHub

- [x] [Create and test Sec8MF update program](https://urbanorg.app.box.com/notes/45544495689)

- [x] [Create and test MFIS update program](https://urbanorg.app.box.com/notes/48593611845)

- [x] [Create and test REAC update program](https://urbanorg.app.box.com/notes/317426216726)

- [x] [Request review and finalize programs](https://urbanorg.app.box.com/notes/295752223069)

- [x] After code has been reviewed and approved, merge branch, sync to `\\sas1\DCData\`, and batch submit all programs updated above.

- [x] Copy full path names of SAS LOG files for **all remote batch submits** into the comments for this issue.  (Example: `\\sas1\DCData\Libraries\PresCat\Prog\Updates\Update_Sec8mf_2018_01.log`)

## Run reports

- [x] Batch submit `\\sas1\DCData\Libraries\PresCat\Prog\Reports\Expiration_rpt.sas`

- [x] Batch submit `\\sas1\DCData\Libraries\PresCat\Prog\Updates\Update_real_property.sas`

- [x] Batch submit `\\sas1\DCData\Libraries\PresCat\Prog\Reports\Real_property_rpt.sas`

- [x] Copy full path names of SAS LOG files for **all remote batch submits** into the comments for this issue.  (Example: `\\sas1\DCData\Libraries\PresCat\Prog\Reports\Expiration_rpt.log`)

## Finalize monthly meeting agenda and Catalog PDF

- [x] Review the output files from each of the updates run above. This includes

- Prog\Updates\Update_Sec8mf_yyyy_mm_subsidy.pdf 

- Prog\Updates\Update_Sec8mf_yyyy_mm_project.pdf

- Prog\Updates\Update_MFIS_yyyy_mm_subsidy.pdf 

[...]

PR #329, opened 2022-12-07
Update #326

dharvey17:
@ptatian Requesting review on the MFIS, SEC8MF and REAC updates. The REAC and Sec8 updates looked fine but it looks like something is going wrong with the MFIS update.
1. Developments that are in the catalog (Samuel Kelsey, Kenyon) appear in the non-match file and,
2. In the subsidy file, the previous old value is now appearing in the "exception" column (see below)
image
g)
[Update_MFIS_2022_10_subsidy_nonmatch.xls](https://github.com/NeighborhoodInfoDC/PresCat/files/10179453/Update_MFIS_2022_10_subsidy_nonmatch.xls)

### DIFF
--- a/Prog/Updates/Update_MFIS_2022_10.sas
+++ b/Prog/Updates/Update_MFIS_2022_10.sas
@@ -1,0 +1,26 @@
+/**************************************************************************
+ Program:  Update_MFIS_2022_10.sas
+ Library:  PresCat
+ Project:  Urban-Greater DC
+ Author:   Donovan Harvey
+ Created:  12/7/22
+ Version:  SAS 9.4
+ Environment:  Local Windows session (desktop)
+ GitHub issue: #326 
+ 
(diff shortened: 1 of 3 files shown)


ptatian:
Hi @dharvey17. Thanks for looking so closely at the output. Both of these things are fine for this update. 
1. The nonmatching subsidies are probably new HUD-insured loans on these existing properties. We currently don't have a way of adding new loans to the subsidy file. Something we should do at some point. 
2. The old values of assisted unit counts actually come from the exception file. The counts are missing in the HUD update file so would be replaced in the Subsidy file on a normal update. But we have the information in the exception file, so they get kept in the end. In other words, the values in the existing Subsidy file get replaced with missing values when the HUD update file is applied, but then get restored using the values in the exception file. So it is all working the way it is supposed to.

Issue #327, opened 2022-11-23
Match TOPA notices for the same properties


ptatian:
This task is to determine how to match TOPA notices across properties, so that all the notices, and related sales data, can be aligned. **The goal would be to add a unique property ID to each notice in the TOPA database.** 

This would likely be done using some combination of the real property SSLs and the MAR address IDs.

PR #331, opened 2022-12-13
Create TOPA_match_notices_property.sas

eburton9:
@ptatian First look at matching TOPA notices to same property.. I wanted to combine address_id, ssl, saledate, owner info, and geos into one dataset. I noticed that for multiple occurrences, the address_id was the same but the ID (notices) would vary, so I sorted by address_id then ID. I have a few questions before continuing, I'll slack you separately but wanted to tag you in this too.


eburton9:
@ptatian I was trying to figure out how to create a property_id from address_id but it did not work out.. I just pushed my code updates based on our conversation on Wednesday and my failed attempt at creating a property_id. I'll return to this in January but let me know if you have any thoughts before then. Thanks!

ptatian:
Thanks, @eburton9 Appreciate the progress you have made! If I have time I might try to move this along. But we can reconnect in January.

eburton9:
@ptatian Hi! I wanted to check if you worked on this while I was out? Before I go back to it. Thanks!

ptatian:
Hi @eburton9 No, I did not. Thanks for checking.

ptatian:
Hi @eburton9. I reviewed your Topa_address_ssl_realprop file and I have an idea for how to finish this up. Here's where we are currently. 

address_id | ID | ssl | Address
-- | -- | -- | --
7 | 344 | 0886    0042 | 701 Florida Avenue NE
7 | 410 | 0886    0042 | 701 Florida Avenue NE
163 | 989 | 5197    0063 | 704 51st Street NE
163 | 1013 | 5197    0063 | 704 51st Street NE
244 | 940 | 3674    0018 | 70 Webster Street NE

The best approach would be to use the address_id as the unique property ID to link notices (in the first two rows, notices 344 and 410). A problem arises for notices that have multiple addresses, however. Resorting the data by ID and address_id: 

address_id | ID | ssl | Address
-- | -- | -- | --
244502 | 10 | 3321    0065 | 124 Webster Street NW
244503 | 10 | 3321    0066 | 126 Webster Street NW
244504 | 10 | 3321    0067 | 128 Webster Street NW
244505 | 10 | 3321    0068 | 130 Webster Street NW

What we want to do, then is use the first address_id as the reference address for the notice. Then when we resort by the reference address_id, all the notices will be lined up.* 

We can do this using some SAS data step magic. Specifically the data step FIRST. operator and the OUTPUT statement. 

First, sort the data set by ID and address_id. 

Next create a data step to read in the sorted data set using the SET statement. But, to use the FIRST. we also need a BY statement for the ID var. 

```
set Topa_address_ssl_realprop_sort;
by id;
```

Within the data step, we want to check if the current row is the first row for a notice (ID). We can use this IF statement like this:

`if first.ID then ... ;  /** not a complete statement **/`

This condition is true if the row is the first in a group of rows with the same ID value (such as address_id=244502 above). When this is true we want to write this observation to the output data set. So the rest of the IF statement is:

`if first.ID then output;`

Using the OUTPUT statement changes the default data step behavior, which is to save every row in the input data set to the output data set. Instead, this IF statement only saves the first obs in each ID group to the output data set. All other obs are discarded. That gives us our ID x address_id crosswalk. For this data step you only need to keep the address_id and ID vars, and I would rename address_id to something like address_id_ref so that we don't confuse it with the other version. 

That concludes the first data step. We would use this by merging back to the other data by ID to add the reference address_id to the data set. We can then sort by the reference address to group the notices. We would want to add this reference address_id to the Google Sheet that CNHED is working with. 

I also think we should save the crosswalk separately as a permanent data set, since we may need it elsewhere. 

Let me know how that sounds and if you have any questions. This approach is discussed [in this SAS paper](https://support.sas.com/resources/papers/proceedings12/041-2012.pdf), which demonstrates the LAST. operator, a counterpart of the FIRST. operator. 

(*The above approach is not foolproof, since separate notices made to the same building might have had different addresses entered. I think that we try to address that by getting to the full list of addresses through the SSL, but it is possible that we might miss some notices that should be matched up. I think that the above approach is about the best we can do for now, however.)

eburton9:
@ptatian thank you so much for this! first and last operator is really good to know about. I updated the program based on instructions and the output looks like what it's suppose to be doing. One question, how do we add the new address_id_ref var to the Google Sheet that CNHED is working with?

ptatian:
That's a great question. I was imagining that we would need to export the data into an Excel or CSV file and then manually copy and paste the new columns into the Google Sheet, making sure they are both properly sorted of course. I don't know if there is a better way. Let me know if you have any other ideas.

eburton9:
I exported the dataset as an excel file earlier.. here it is in box [but](https://urbanorg.box.com/s/q1xx5homsgqxh8kbbiplx8l4eypqwidp) But I'm realizing that there's around 7,000 obs in this excel compared to 1,745 in CNHED's spreadsheet because we parsed out the addresses from CNHED to add address_id and ssl from the geocoder in the previous program.

ptatian:
But is that the data set you created using the first. with the data step? That should have the same number of rows as the CNHED spreadsheet, that is, one obs per ID value.

eburton9:
@ptatian I see, yes. I merged the wrong way yesterday. I thought if I changed the dataset from the first data step, the address_id x id crosswalk (only 1693 obs), to the right side of the merge than it would output the right dataset's number of obs but that's not the case. I googled for 20-30 min but I can't figure out how to a merge where the dataset with the smallest number of obs would override the dataset with the larger # of obs.

ptatian:
Thanks, @eburton9 . What can you tell about the 6 obs that are missing? Do you see any reason why they would be missing?

eburton9:
@ptatian I compared the address_id and id crosswalk to the original csv we imported into sas and i identifed the 6 IDs by comparing them and they are all missing a full address. The following are the IDs with what the address on CNHED's spreadsheet is, 417: 4A03; 467: LO; 471: LO; 482: LO; 528: LO; 1542: Holmstead Place Apartments. I think the LO ones are a mistake on CNHED's part because LO is an initial used in the verified by column.

eburton9:
@ptatian Sorry - I went ahead and checked CNHED's most [up to date spreadsheet](https://docs.google.com/spreadsheets/d/1xUxa8vBC8kZ3BLSiY8hv3LbrdvSGuF6SZfnfwzCMoIQ/edit?usp=sharing) and the LO addresses now have real ones (IDS 467,471,482,528) but ID 417 is still only 4A03 on the spreadsheet and ID 1542 is still Holmstead Place Apartments. They've also added 46 new notices, so probably worth re uploading the csv and re running previous program at some point.

eburton9:
@ptatian Yes I do - could you look at my most recent code first? I change the merge to a data set step to have the numbers of obs be 1693 (crosswalk) rather than the 7000 from parsing out addresses. The output looked right to me but want to make sure first.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,101 @@
...
+%File_info( data=PresCat.TOPA_database, printobs=5 ) /** 1699 obs**/
+
+/** merge address_id ssl and real_prop info **/
+
+data Topa_address_ssl_realprop;
+  merge
+    PresCat.TOPA_SSL (keep=id address_id ssl)
+    PresCat.TOPA_realprop (keep=id SALEDATE offer_sale_date CASD_date ADDRESS1 ADDRESS3 Ownername_full days_notice_to_sale ui_proptype Anc2012 cluster2017 Geo2020 GeoBg2020 GeoBlk2020 Psa2012 VoterPre2012 Ward2022 Ward2012)
+    PresCat.TOPA_addresses (keep =id address);
+run;
```
ptatian, line 35: @eburton9 You should have a BY statement whenever you use MERGE. Otherwise, the observations just match up based on their order in the respective data sets, which might not be correct. Did you want to match by ID?

eburton9:
@ptatian Yes- added by id to merge

ptatian:
Thanks, I won't be able to check this again until tomorrow morning.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -61,7 +62,6 @@ proc export data=Topa_id_x_address
 run;
 
 /** address_id_ref as property id **/
-
 data Topa_addressid_merge; 
   set Topa_address_ssl_realprop_sort; 
   set Topa_id_x_address;
```
ptatian: @eburton9 This is similar to doing a merge without a BY statement. Having two SETs will merge the based on the order of the rows in the respective data sets, but they won't necessarily match by ID. Normally you should not try to merge this way. 
But also I don't think it is necessary to merge to the real property data for what we need here.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,100 @@
...
+%File_info( data=PresCat.TOPA_realprop, printobs=5 ) /** some IDs don't have real_prop info**/
+%File_info( data=PresCat.TOPA_addresses, printobs=5 )
+%File_info( data=PresCat.TOPA_database, printobs=5 ) /** 1699 obs**/
+
+/** merge address_id ssl and real_prop info **/
+
+data Topa_address_ssl_realprop;
+  merge
+    PresCat.TOPA_SSL (keep=id address_id ssl in=in_ssl)
+    PresCat.TOPA_addresses (keep =id address casd_date offer_sale_date in=in_addresses);
```
ptatian, line 33: I took out the real property data set here, which I don't think we need for this step. And doing 3-way merges in a data step can lead to unpredictable results.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,100 @@
...
+  by id; 
+  /** List in LOG notices that are not in both data sets **/
+  if not ( in_ssl and in_addresses ) then put in_ssl= in_addresses= id= ssl= address_id= address=;
```
ptatian, line 36: This code lets us see in the LOG which notices are not in both data sets.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,100 @@
...
+
+/** sorting by id then address id**/
+proc sort data=Topa_address_ssl_realprop out=Topa_address_ssl_realprop_sort;
+  by id address_id;
+run;
+
+%File_info( data=Topa_address_ssl_realprop_sort, printobs=5 )
+
+/** create ID (notice) x address_id crosswalk **/
+data Topa_id_x_address_1;
```
ptatian, line 49: Renamed the output data set so that it can be used the next data step I added.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,100 @@
...
+run;
+
+%File_info( data=Topa_address_ssl_realprop_sort, printobs=5 )
+
+/** create ID (notice) x address_id crosswalk **/
+data Topa_id_x_address_1; 
+  set Topa_address_ssl_realprop_sort;
+  by id; 
+  if first.id then output; 
+  keep address_id id casd_date offer_sale_date;
```
ptatian, line 53: Added the two date vars here, which I thought might be helpful to see in the output file.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,98 @@
...
+  rename address_id=address_id_ref;
+run; 
+
+/** Fill in missing ID numbers to match original TOPA database **/
+data Topa_id_x_address; 
+  merge 
+    Topa_id_x_address_1
+    Prescat.Topa_database (keep=id all_street_addresses address_for_mapping);
+  by id;
+run;
```
ptatian, line 63: I added this data step to fill in the missing IDs from the original TOPA database. By doing a merge by ID, any missing IDs in the Topa_id_x_address_1 will be added to the output data set. Also, I added some vars from the database that I though would be useful in examining the results.

Prog/AddNew/TOPA_match_notices_property.sas
```diff
@@ -0,0 +1,98 @@
...
+run; 
+ 
+%File_info( data=Topa_id_x_address, printobs=5 ) /** 1693 obs**/
+
+title2 '** Notices with missing address_id_ref **';
+proc print data=Topa_id_x_address;
+  where missing( address_id_ref );
+  id id;
+  var address_id_ref all_street_addresses address_for_mapping;
+run;
```
ptatian, line 72: This proc print will list any IDs with missing address_id_ref, which will be helpful for identifying records that might need further cleaning.

ptatian:
Hi @eburton9 What you had was good, but I made a few changes, as noted above. The export file now has all 1,699 IDs in it, so it should be easier to match up with the Google Sheet. Please let me know if you have any questions. 
I would like to keep this issue open for now until we review what we have with the project team and see if any other changes are needed. Thanks!

eburton9:
Thank you! @ptatian I added the address_id_ref variable to the google sheet.

eburton9:
frequency of notices -> figure out notices in same property that are only a few months or less apart 
use variation of recent crosswalk to use first, last, and retain statement (keeps info from previous obs so keep first date and calculate number of days between the two)

eburton9:
@ptatian Committed my attempt to find the number of days between offer of sale date by address_if_ref, I'm understanding retain, last, and first (or so I thought), but can't figure out how to calculate the number of days between the first obs of an addresse_if_ref and the second. (other things are definitely wrong too)

ptatian:
Hi @eburton9 Thanks for taking a pass at this. I think I see the confusion. The new var for the previous date is the one you want to retain, so that you can find the difference between the retained date and the current notice date. In your code: 

```
data Topa_notice_freq; 
  set Topa_id_x_address; 
  by address_id_ref; 
  retain offer_sale_date address_id_ref;
  if first.address_id_ref then fr_offer_sale_date=offer_sale_date; /** I'm trying to create new var that is the first date of notice**/
  days_btwn_notices = fr_offer_sale_date - offer_sale_date; /** Trying to subtract first notice date to the next obs **/
  if last.address_id_ref then output; /** Trying to just keep the same info if only one address_id_ref**/
  run;
```

You want to retain **fr_offer_sale_date**, not the two vars you have now. This var will keep the notice date from the previous notice. Now let's think about the logic. 

If this is the first notice for the **address_id_ref**, then **days_btwn_notices** should be missing (.) because there was no previous notice for this property. Since **days_btwn_notices** is a new var, and not a retained var, you actually don't need to do anything explicit for this. The value will automatically be set to missing. But, in addition, we want to set the value of **fr_offer_sale_date** to be equal to the current **offer_sale_date**, as you are doing.  So the `if first.address_id_ref then ...` statement you have is perfectly fine. 

If this is **not** the first notice, however, then we want to calculate **days_btwn_notices**, as you do above. (Although, you have the dates reversed because **offer_sale_date** is going to be later than **fr_offer_sale_date**. You'll get a negative value the way you have it now.) The point of retaining **fr_offer_sale_date** is that the date of the previous notice will be carried forward to the current notice. Then you can calculate the difference between the two dates. Once you have calculated **days_btwn_notices**, however, you want to set **fr_offer_sale_date** equal to **offer_sale_date**, so that you can use it in the _next_ iteration of the data step. It's important to do those statements in that sequence.

You can keep it for now as a check, but for the final code you will want to drop **fr_offer_sale_date** from the output data set. 

That should be it. You don't need any last. condition and should not use an output statement because we want to save every obs in the original data set. 

Let me know if that makes sense or if you have any questions.

eburton9:
Thank you, @ptatian. The output looks correct now. However, when looking through it, I noticed an instance where the offer sale date was the same date for two notices of the same building but the days between notice variable created is 0 which blends in with the rest of the 0s that really should be NAs instead (only one notice per address_id_ref). I tried to create days between notice asmissing if it's the first address_id_ref to distinguish but it returns zeros still. Any other ideas or is this not a concern?

ptatian:
Hi @eburton9 Thanks for checking on this. I think having the 0 days between notices is fine. It doesn't seem like there's a way to distinguish the two notices with the same date. They could be duplicates or perhaps a data entry error. We'll delete one of them when we do the clean up of duplicate notices. 
If the data looks OK to you, can you add the days between notices column to the Google Sheet?

eburton9:
@ptatian Just did, the sheet should be sorted by the address id ref to make any sense though

ptatian:
Hi @eburton9. I reviewed and made some changes to the code for the TOPA database. Main change was merging the SSL-address crosswalk a second time to fill out a more complete list of addresses for each property, rather than relying only on the addresses that were entered on the notice. I created a new TOPA_addresses data set that will replace the one we have currently. This dataset also has all the geography information on it, so we don't need to save the separate TOPA_geocoded data set. 

I also cleaned up the variable names on the TOPA_database by adding a new row to a new input CSV file (called TOPA_DOPA_5+_with_var_names.csv) that has new variable names. The variable names are similar those created by Proc Input by default but I removed things like the multiple underscores and the extra information (Y/N, usually from DHCD, etc.) I left the original first row as row 2, and the data now start on row 3. I added the `datarow=3;` statement to Proc Import so that it will skip the second row.

I used the original column headings to create variable labels, which will be helpful later when we re-export the data. The new variable labels can be found in PresCat\Raw\TOPA\TOPA_DOPA_5+_variable_labels.txt on SAS1, which is read into the SAS program using %include. In that same SAS1 folder, you'll find an Excel file that shows how I generated the label statements--which is a useful trick to know. I copied the first two rows from the input CSV file and then used the Transpose option when copying and pasting them to a new workbook (Paste Special > Paste Special > check the Transpose box and click OK) to switch them from columns to rows. Then I wrote an Excel formula to create label assignments, and copied and pasted them into a plain text file. (I decided to keep them as a separate file for neatness, but I could have just pasted them directly into the .SAS program.) 

A few other minor changes, such as labeling variables. I also renamed the new variables we are creating to start with u_ so that they are easily distinguishable from the other variables, which again will be helpful later. I also added some code to address the few entries without valid addresses, such as by using the mapping address the students entered, although there remains one row (3536 Carter St NW) that we can't fix without more information. 

---

For **next steps** to finish this issue, as I said I'd like you to download the latest version of CNHED TOPA notice database from Google, and then use it as the new input file for TOPA_data_to_DC_data.sas. You should delete the address_id_ref and Days between Notices columns from the CSV file, since we'll be redoing these later. You'll need to add the variable name row, like I did on the older file. Also, please double-check the columns because I think Martha may have added one or more new ones. I can review once you've done that and then we can finalize TOPA_data_to_DC_data.sas and those data sets.

For the TOPA_match_notices_property.sas program, we will finish this as a separate issue because we need the data sets created by TOPA_data_to_DC_data.sas for this program, but we can't finalize those data sets until we merge the branch. So it is better to have them in separate branches. You can just leave the program as is for now (along with the Topa_quick_analysis.sas program that I created). Once we merge branch enhance-#327 into main, we can create a new branch and continue working on TOPA_match_notices_property.sas under a new issue. 

Also, I think it would be better move these programs into their own folder, since we are going to have more of them, so would you please move the TOPA programs to Prog\TOPA. I'd probably do that first, in case you have to make any changes to the programs, which will then be easier to track. 

Thanks, Elizabeth please let me know if you have any questions.

Prog/AddNew/TOPA_data_to_DC_data.sas
```diff
@@ -132,23 +162,80 @@ proc sql noprint;
...
 	TOPA_SSL.GeoBg2020, TOPA_SSL.GeoBlk2020, TOPA_SSL.Psa2012,
 	TOPA_SSL.VoterPre2012, TOPA_SSL.Ward2022, TOPA_SSL.Ward2012,
     RealProp.SSL, RealProp.SALEPRICE, RealProp.saleprice_prev, RealProp.SALEDATE, RealProp.saledate_prev, RealProp.Ownername_full, RealProp.ownername_full_prev, RealProp.ui_proptype,
-	RealProp.address1, RealProp.address2, RealProp.address3, RealProp.address1_prev, RealProp.address2_prev, RealProp.address3_prev  
+	RealProp.address1, RealProp.address2, RealProp.address3, RealProp.address1_prev, RealProp.address2_prev, RealProp.address3_prev, RealProp.premiseadd, RealProp.hstd_code  
     from TOPA_SSL (where=(not(missing(SSL)))) as TOPA_SSL
       left join RealProp.Sales_master as realprop    /** Left join = only keep obs that are in TOPA_geocoded **/
   on TOPA_SSL.SSL = realprop.SSL   /** This is the condition you are matching on **/
-  where SALEDATE > max(CASD_date, offer_sale_date) /** obs where sale date is after the later of CASD data and offer of sale data **/
+  where SALEDATE > max(u_CASD_date, u_offer_sale_date) /** obs where sale date is after the later of CASD data and offer of sale data **/
```
ptatian, line 176: @eburton9 Sorry, I forgot to mention this in my earlier message, but let's just take out this where clause for the real property data. We want to see prior sales, too, and we can figure out the sequencing using the method that I shared in the email I sent on Friday. Thanks!

eburton9:
@ptatian I went ahead and ran this file and received a lot of errors. I'll go through them more in depth tomorrow but one that I didn't know how to address: ERROR: [DC_mar_geocode] Invalid keyword GEO2020 found in KEEP_GEO= parameter. Later, there seems to be an issue with finding ward2022 var too.. so wondering if there's an issue with 2020 geographies. I think I helped Rob update those though and this program ran without errors earlier.

ptatian:
Hi @eburton9. The code worked for me. Do you have the MAR repo on your local/virtual desktop? If so, make sure you pull the latest updates for the main branch to the local repo and try submitting the program again.

eburton9:
@ptatian Yup that was the issue, all good now.

ptatian:
Thanks, Elizabeth. Let me know when you are ready for me to review.
________________________________

eburton9:
@ptatian whoops sorry, ready for review!

PR #333, opened 2023-01-25
Sec8MF, MFIS and REAC updates

dharvey17:
@ptatian Ready for review; flagging that MFIS_subsidy_nonmatch file includes a number of properties that are definitely in the Catalog (Kenyon, Samuel Kelsey, Edgewood Terrace, etc.)

### DIFF
--- a/Prog/Updates/Update_MFIS_2022_12.sas
+++ b/Prog/Updates/Update_MFIS_2022_12.sas
@@ -1,0 +1,26 @@
+/**************************************************************************
+ Program:  Update_MFIS_2022_12.sas
+ Library:  PresCat
+ Project:  Urban-Greater DC
+ Author:   Donovan Harvey
+ Created:  1/25/23
+ Version:  SAS 9.4
+ Environment:  Local Windows session (desktop)
+ GitHub issue:  Update-#332
+ 
+ Description:  Update Preservation Catalog with latest 
+ HUD MFIS update file.
+
+ Modifications:
+**************************************************************************/
+
+%include "\\sas1\DCdata\SAS\Inc\StdLocal.sas";
+
+** Define libraries **;
+%DCData_lib( PresCat )
+%DCData_lib( HUD )
+
+
+%Update_MFIS( Update_file=MFIS_2022_12 )
+
+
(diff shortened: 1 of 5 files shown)


Prog/Updates/Update_Sec8mf_2022_12.sas
```diff
@@ -0,0 +1,25 @@
...
+**************************************************************************/
+
+%include "\\sas1\DCdata\SAS\Inc\StdLocal.sas";
+
+** Define libraries **;
+%DCData_lib( PresCat )
+%DCData_lib( HUD )
+
+options nominoperator;
+%Update_Sec8mf( Update_file=Sec8mf_2022_12 )
```
ptatian, line 24: This looks good but there is a Section 8 project that appears not to be in our Catalog. Strange. I will create a separate issue to add it.

Prog/Updates/Update_MFIS_2022_12.sas
```diff
@@ -0,0 +1,26 @@
...
+**************************************************************************/
+
+%include "\\sas1\DCdata\SAS\Inc\StdLocal.sas";
+
+** Define libraries **;
+%DCData_lib( PresCat )
+%DCData_lib( HUD )
+
+
+%Update_MFIS( Update_file=MFIS_2022_12 )
```
ptatian, line 24: As you noted, there are a few loans here that are not matching up with their respective projects. That is because the subsidy record has not been created for these loans. I will create a separate issue for this update, which we should complete before finalizing this issue.

ptatian (changes requested):
Hi @dharvey17 A few things we should do before finalizing, to fix some of the non-matches. In addition to MFIS, we have nonmatches in REAC and a section 8 project that seems to be missing from the Catalog. I will create separate issues for the MFIS and Sec8MF nonmatches, but the REAC one can be addressed within this issue by following the instructions in my comment.

Prog/Updates/Update_REAC_2023_01.sas
```diff
@@ -0,0 +1,25 @@
...
+**************************************************************************/
+
+%include "\\sas1\DCdata\SAS\Inc\StdLocal.sas";
+
+** Define libraries **;
+%DCData_lib( PresCat )
+%DCData_lib( HUD )
+
+
+%Update_REAC( Update_file=REAC_2023_01 )
```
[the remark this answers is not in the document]
dharvey17: Done. The Gregory and Woodmont Crossing no longer show up in the nonmatching records output (I see your point on SAS being satisfying)

ptatian (approved):
Hi @dharvey17 The REAC update looks good! Glad you got some satisfaction from it. This is OK to finalize.

Issue #337, opened 2023-03-04
Add unique property (address) ID to TOPA database


ptatian:
Repeat analysis steps from issues #321 and #327, adapting them for the new clean TOPA database created in issue #336. 

Final result should be new version of the TOPA database data set created in #336 that has Urban provided address_id_ref (now lets call it u_address_id_ref) and days between notices (u_days_between_notices). But also add another variable, u_days_from_first_notice, which will be the number of days from the first notice for the property to the current notice.

ptatian:
To expand more on the description above, using [one of the examples](https://docs.google.com/document/d/1nYwyYNkVlMNjTB1iWIv53O2X1ksMtQQt/edit?usp=sharing&ouid=101012864185545505029&rtpof=true&sd=true) I created with the Quick_analysis.sas program. I didn't do this in my program, but we should limit both sale and notice data to 2006 to 2020 (inclusive). If we sort the date by u_address_id_ref and descending date, we have:

**Property ID (u_address_id_ref)=5142**

Date | Type of activity | TOPA notice ID
-- | -- | --
05/20/2020 | NOTICE OF SALE | 1452
05/13/2014 | SALE | .
03/06/2014 | NOTICE OF SALE | 766
08/01/2013 | NOTICE OF SALE | 668
11/07/2006 | SALE | .
10/06/2006 | NOTICE OF SALE | 37

The rule will be to tag the first notice before a sale as u_dedup_notice = 1. Also, I think we need another flag which is u_notice_with_sale, which = 1 for notices that had a sale by 2020, ,or = 0 if not. 

In a data step using the above sorted data, if you encounter a SALE record, then you can create a flag variable = 1 and **retain** that variable to the next record. If the next record is a NOTICE OF SALE for the same property, then that notice would be flagged u_dedup_notice = 1 and u_notice_with_sale = 1. Other notices would be flagged as u_dedup_notice = 0  and u_notice_with_sale = 0. An exception is where a notice is the first observation for a property, like 5/20/2020, which would be flagged as u_dedup_notice = 1 but u_notice_with_sale = 0.

In addition, we want to carry forward the sales data listed in the [data analysis plan 5.a](https://docs.google.com/document/d/1hnet3wDgDR876SYBmAefSgXPnmcH6Ybyqj7bJ3xrHoo/edit?usp=sharing) to the dedup notice record. More retain statements. And calculate the number of days to the sale u_days_from_dedup_notice_to_sale. 

[...]

PR #349, opened 2023-03-23
Update #337

eburton9:
@ptatian Here is the first part of Issue #337 - I updated the u_address_id_ref and the u_days_between_notices. I also took your code from the Combo data set in TOPA quick analysis and added a condition to only keep noticies and sales between 2006 and 2020. I've been working on the flag u_dedup_notice, is it correct to use the first operator for u_address_id_ref (property id) and assign a flag? Then, not sure how to have SAS say to assign something in the next obersation after temp_flag =1?

For the flag u_notice_with_sale , is it "which = 1 for notices that had a sale by 2020, ,or = 0 if not" or "1 if the notice was the last notice before a property sale; = 0 otherwise"? There were different definitions in your first and second comment.

For the u_days_from_dedup_notice_to_sale, is the dedupilcated notice based on the u_dedup_notice flag created above? 

One last small question for now, I see you added a check for the notices with missing u_address_id_ref, do I need to do anything about the missing property ids right now?

### DIFF
--- a/Prog/TOPA/TOPA_match_notices_property.sas
+++ b/Prog/TOPA/TOPA_match_notices_property.sas
@@ -72,6 +72,7 @@ data Sales_by_property;
   merge
     Prescat.Topa_realprop Topa_id_x_address;
   by id;
+  if missing( ssl ) then delete;
 run;
 
 proc sort data=Sales_by_property out=Sales_by_property_nodup nodupkey;
@@ -102,9 +103,14 @@ data Combo;
 	   rename=(saledate=ref_date));
 	by u_address_id_ref descending ref_date;
 
+	length desc $ 40;
+
 	if is_notice then desc = "NOTICE OF SALE";
 	else desc = "SALE";
-  	where ref_date between '01Jan2006'd and '31dec2020'd; /**Limit sale and notice data to 2006-2020**/
+
+    /**Limit sale and notice data to 2006-2020 and do not use obs with missing address or date **/
+    where ref_date between '01Jan2006'd and '31dec2020'd 
+	  and not( missing( u_address_id_ref ) or missing( ref_date ) );
 
 run;
 
@@ -125,6 +131,15 @@ data Topa_notice_flag;
   if ID and not(temp1) then do; temp1=1; u_dedup_notice=1;end;else u_dedup_notice=0; 
 /*  drop temp1;*/
 run; 
+
+/** Temporary Proc Print for checking results **/
+proc print data=Topa_notice_flag;
+  where u_address_id_ref=5142;
+  by u_address_id_ref;
+  id ref_date;
+  var desc id temp1 u_dedup_notice;
+run;
+
    
 %File_info( data=Topa_notice_flag, printobs=50) /** 4050 obs **/
 
--- a/Prog/TOPA/TOPA_match_notices_property.sas
+++ b/Prog/TOPA/TOPA_match_notices_property.sas
@@ -16,30 +16,32 @@
 
 ** Define libraries **;
 %DCData_lib( PresCat )
-%DCData_lib( DHCD )
-%DCData_lib( MAR )
-%DCData_lib( RealProp )
 
 %File_info( data=PresCat.TOPA_SSL, printobs=5 )
 %File_info( data=PresCat.TOPA_realprop, printobs=5 ) /** some IDs don't have real_prop info**/
 %File_info( data=PresCat.TOPA_addresses, printobs=5 )
-%File_info( data=PresCat.TOPA_database, printobs=5 ) /** 1699 obs**/
+%File_info( data=PresCat.TOPA_database, printobs=5 ) 
 
+/*create flags (u_dedup_notice & u_notice_with_sale )*/
+/*create u_days_between_notices & u_days_from_dedup_notice_to_sale*/
+/*sales data listed in the data analysis plan 5.a*/
+
+/** Creating an ID for each property **/
 /** sorting by id then address id**/
-proc sort data=Topa_addresses_full out=Topa_addresses_full_sort;
+proc sort data=Prescat.Topa_ssl out=Topa_ssl_sort;
   by id address_id;
 run;
 
-%File_info( data=Topa_address_ssl_realprop_sort, printobs=5 )
-
 /** create ID (notice) x address_id crosswalk **/
 data Topa_id_x_address_1; 
-  set Topa_addresses_full_sort;
+  set Topa_ssl_sort;
   by id; 
   if first.id then output; 
-  *keep address_id id casd_date offer_sale_date; 
-  rename address_id=address_id_ref;
+  rename address_id=u_address_id_ref;
 run; 
+
+%File_info( data=Topa_id_x_address_1, printobs=5 ) /** 1699 obs**/
+
 
 /** Fill in missing ID numbers to match original TOPA database **/
 data Topa_id_x_address; 
@@ -49,13 +51,13 @@ data Topa_id_x_address;
   by id;
 run; 
  
-%File_info( data=Topa_id_x_address, printobs=5 ) /** 1699 obs**/
+%File_info( data=Topa_id_x_address, printobs=5 ) /** 1740 obs**/
 
-title2 '** Notices with missing address_id_ref **';
+title2 '** Notices with missing u_address_id_ref **';
 proc print data=Topa_id_x_address;
-  where missing( address_id_ref );
+  where missing( u_address_id_ref );
   id id;
-  var address_id_ref all_street_addresses address_for_mapping;
+  var u_address_id_ref all_street_addresses address_for_mapping ssl;
 run;
 title2;
 
@@ -65,21 +67,77 @@ title2;
 /*    replace;*/
 /*run;*/
 
+/*Add vars from real_prop, combine with TOPA_id_x_address created above, clean up variables, and limit sale/notice data to 2006-2020*/
+data Sales_by_property;
+  merge
+    Prescat.Topa_realprop Topa_id_x_address;
+  by id;
+run;
+
+proc sort data=Sales_by_property out=Sales_by_property_nodup nodupkey;
+  by u_address_id_ref saledate;
+run;
+
+%File_info( data=Sales_by_property_nodup, printobs=5 ) /** 5009 obs**/
+
 proc sort data=Topa_id_x_address;
-  by address_id_ref id;
+  by u_address_id_ref u_offer_sale_date id;
+run;
+
+data Combo;
+  set 
+    Topa_id_x_address 
+    (keep=u_address_id_ref id u_offer_sale_date
+     rename=(u_offer_sale_date=ref_date)
+     in=is_notice)
+    Sales_by_property_nodup
+	  (keep=u_address_id_ref saledate saleprice ownername_full ui_proptype
+	   rename=(saledate=ref_date));
+	by u_address_id_ref ref_date;
+
+	if is_notice then desc = "NOTICE OF SALE";
+	else desc = "SALE";
+  	where ref_date between '01Jan2006'd and '31dec2020'd; /**Limit sale and notice data to 2006-2020**/
+
+run;
+
+%File_info( data=Combo, printobs=5 ) /** 4050 obs**/
+
+/*Create flags (u_dedup_notice & u_notice_with_sale)*/
+proc sort data=Combo;
+  by u_address_id_ref descending ref_date;
+run;
+
+%File_info( data=Combo, printobs=5 ) /** 4050 obs**/
+
+data Topa_notice_flag; 
+  set Combo;  
+  if desc="SALE" then temp_flag=1; 
+  else temp_flag=0; 
+  retain temp_flag; 
+  if desc="NOTICE OF SALE" then u_dedup_notice = 1; 
+  else u_dedup_notice = 0;
+/*  if first.id then temp_flag=1; */
+/*  retain temp_flag;*/
+
+run; 
+   
+%File_info( data=Topa_notice_flag) /** 4050 obs**/
+
+/*Create u_days_between_notices & u_days_from_dedup_notice_to_sale*/
+proc sort data=Topa_id_x_address;
+  by u_address_id_ref id;
 run;
 
 data Topa_notice_freq; 
   set Topa_id_x_address; 
-  by address_id_ref; 
-  if first.address_id_ref then fr_offer_sale_date=offer_sale_date;
-  retain fr_offer_sale_date;
-  days_btwn_notices = offer_sale_date - fr_offer_sale_date;  
-  fr_offer_sale_date = offer_sale_date;
-  format fr_offer_sale_date MMDDYY10.;
+  by u_address_id_ref; 
+  if first.u_address_id_ref then u_fr_offer_sale_date=u_offer_sale_date;
+  retain u_fr_offer_sale_date;
+  u_days_between_notices = u_offer_sale_date - u_fr_offer_sale_date;  
+  u_fr_offer_sale_date = u_offer_sale_date;
+  format u_fr_offer_sale_date MMDDYY10.;
   run;
-
-%File_info( data=Topa_notice_freq, printobs=5 ) /** 1699 obs**/
 
 proc sort data=Topa_notice_freq; 
   by id; 
--- a/Prog/TOPA/TOPA_match_notices_property.sas
+++ b/Prog/TOPA/TOPA_match_notices_property.sas
@@ -96,12 +96,12 @@ data Combo;
   set 
     TOPA_by_property 
     (keep=u_address_id_ref id u_offer_sale_date
-     rename=(u_offer_sale_date=ref_date)
+     rename=(u_offer_sale_date=u_ref_date)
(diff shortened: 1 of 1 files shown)


ptatian:
Hi @eburton9  Thanks for moving this ahead. To respond to your questions.

> is it correct to use the first operator for u_address_id_ref (property id) and assign a flag? Then, not sure how to have SAS say to assign something in the next obersation after temp_flag =1?

You can use first.u_address_id to see when you are starting with a new property. That would be the last notice or sale record associated with the property (within the 2006-2020 window).

Then you want to carry the information from temp_flag to the next observation. If you include temp_flag in a retain statement, then the value will be held over to the next observation.

> For the flag u_notice_with_sale , is it "which = 1 for notices that had a sale by 2020, ,or = 0 if not" or "1 if the notice was the last notice before a property sale; = 0 otherwise"? There were different definitions in your first and second comment.

Sorry if that was confusing. u_notice_with_sale indicates that there was a sale that happened after the notice. Using the example I had in the issue, here's what that looks like.

Date | Type of activity | TOPA notice ID | u_dedup_notice | u_days_from_dedup_notice_to_sale  | u_notice_with_sale
-- | -- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1452 | 1 | . | 0
05/13/2014 | SALE | . | n/a | n/a | n/a
03/06/2014 | NOTICE OF SALE | 766 | 1 | _xxx_ | 1
08/01/2013 | NOTICE OF SALE | 668 | 0 | . | 0
11/07/2006 | SALE | . | n/a | n/a | n/a
10/06/2006 | NOTICE OF SALE | 37 | 1 | _xxx_ | 1

(where _xxx_ = the number of days)

I should have also said that, in the output data set, we are only going to save the NOTICE OF SALE observations, not the SALE observations. The data from the SALE (sale date, owner, etc.) needs to be added as to the records where u_notice_with_sale=1. So those values also need to be RETAIN'ed from the sale obs to the notice obs. 

Now, one thing to be aware of with RETAIN--it only works with new variables that are created in the data step. For variables in existing data sets, those values are always updated when a new observation is read from the input data set. So, you will have to create new versions of the sale variables that will be retained and then assign values to them from the SALES record. Maybe to stick with our naming conventions, call the new variables u_saledate, u_ownername_full, etc. You can then drop the original sales variables from the real property data from the output data set and keep the new variables. 

> For the u_days_from_dedup_notice_to_sale, is the dedupilcated notice based on the u_dedup_notice flag created above?

Yes.

> One last small question for now, I see you added a check for the notices with missing u_address_id_ref, do I need to do anything about the missing property ids right now?

I don't think so. The new TOPA_addresses data set should not have any missing u_address_id_ref. But I always like to _program defensively_, and anticipate anything that might go wrong. If we do see any missing values, we should deal with that.

Let me know if you have more questions, or if it would be helpful to talk any of this over. Thanks!

Prog/TOPA/TOPA_match_notices_property.sas
```diff
@@ -16,31 +16,33 @@
...
+%File_info( data=PresCat.TOPA_database, printobs=5 ) 
 
+/*create flags (u_dedup_notice & u_notice_with_sale )*/
+/*create u_days_between_notices & u_days_from_dedup_notice_to_sale*/
+/*sales data listed in the data analysis plan 5.a*/
+
+/** Creating an ID for each property **/
 /** sorting by id then address id**/
-proc sort data=Topa_addresses_full out=Topa_addresses_full_sort;
+proc sort data=Prescat.Topa_ssl out=Topa_ssl_sort;
```
ptatian, line 31: @eburton9 You should use the Prescat.TOPA_addresses data set here.

eburton9:
@ptatian Committed my updates for today. I think I got the dedup_notice working.. I started on the u_notice_with_sale but I'm only able to figure it out to have the first notice that doesn't have a sale be 0, not the one following it or for ones with sales, the subsquent notices that are duplicates to be 0. I'm sure I'm doing it wrong. One other thing, there are two address_ids missing after doing the check: notice 83 and notice 415.

ptatian:
Thanks for the update, @eburton9. I'll review what you did and get back to you next week.

ptatian:
Hi @eburton9. Thanks for the first pass at this. Data step programming can be tricky stuff, so let's take it one piece at a time. Starting with the u_dedup_notice variable. It isn't quite right yet. I added a Proc Print to the program that will show what things look like for the example I included above, so that we can see how that compares to our current code. Here's what it looks like after the Topa_notice_flag data step. 

**DC MAR address ID=5142**

ref_date | desc | ID | temp1 | u_dedup_notice 
-- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 
05/13/2014 | SALE | . | 1 | 0 
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 
11/07/2006 | SALE | . | 1 | 0 
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 

We want u_dedup_notice = 1 for notice IDs 1449, 764, and 37, and = 0 otherwise. 1449 because it is the first obs (in this sort order) for the u_address_id_ref AND it is a NOTICE OF SALE. 764 and 37 because they are the last notices before a SALE (that is, they immediately follow a SALE in this sort order). 

The temp1 flag you have now is always 1 in this case, so not especially useful. It looks like you may be trying to use it to flag the first obs for the property, but it isn't necessary to create a new var for that because that is what first.u_address_id_ref tells you. Think of first.u_address_id_ref like another var, but one that isn't saved to the data set. If you were to actually see it within the data step, here's what it would look like. 

ref_date | desc | ID | temp1 | u_dedup_notice | first.u_address_id_ref
-- | -- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 | 1
05/13/2014 | SALE | . | 1 | 0 | 0
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 | 0
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 | 0
11/07/2006 | SALE | . | 1 | 0 | 0
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 | 0

Back to u_dedup_notice. **First, we want to write an IF statement to set u_dedup_notice to 1 if the observation is the first one for u_address_id_ref and it is a notice of sale.** You do not need to create any additional variables to write this IF condition; all the information we need is in the data we have. (BTW, please use the desc variable to distinguish between notices of sale and sales. Some sales are missing sales price, for instance.) 

Once you have written that IF statement, we need to think about what happens if that condition isn't true. That is, we need to check whether it meets the other condition for u_dedup_notice to be set to 1: it is a NOTICE OF SALE that was preceded by a SALE for the same u_address_id_ref. 

Now, if you are looking at a NOTICE OF SALE observation, we don't have anything currently that would tell us that it was preceded by a SALE. So we need to create a variable with that information. **Think about how you would create that variable using RETAIN.** 

I also want to mention that all this can be done in a single data step. 

Let's focus on getting this far first and then we can build on that. If it would be helpful to talk it through, let's find some time to do that. Thanks!

ptatian:
Hi @eburton9. Thanks for the first pass at this. Data step programming can be tricky stuff, so let's take it one piece at a time. Starting with the u_dedup_notice variable. It isn't quite right yet. I added a Proc Print to the program that will show what things look like for the example I included above, so that we can see how that compares to our current code. Here's what it looks like after the Topa_notice_flag data step. 

**DC MAR address ID=5142**

ref_date | desc | ID | temp1 | u_dedup_notice 
-- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 
05/13/2014 | SALE | . | 1 | 0 
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 
11/07/2006 | SALE | . | 1 | 0 
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 

We want u_dedup_notice = 1 for notice IDs 1449, 764, and 37, and = 0 otherwise. 1449 because it is the first obs (in this sort order) for the u_address_id_ref AND it is a NOTICE OF SALE. 764 and 37 because they are the last notices before a SALE (that is, they immediately follow a SALE in this sort order). 

The temp1 flag you have now is always 1 in this case, so not especially useful. It looks like you may be trying to use it to flag the first obs for the property, but it isn't necessary to create a new var for that because that is what first.u_address_id_ref tells you. Think of first.u_address_id_ref like another var, but one that isn't saved to the data set. If you were to actually see it within the data step, here's what it would look like. 

ref_date | desc | ID | temp1 | u_dedup_notice | first.u_address_id_ref
-- | -- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 | 1
05/13/2014 | SALE | . | 1 | 0 | 0
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 | 0
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 | 0
11/07/2006 | SALE | . | 1 | 0 | 0
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 | 0

Back to u_dedup_notice. **First, we want to write an IF statement to set u_dedup_notice to 1 if the observation is the first one for u_address_id_ref and it is a notice of sale.** You do not need to create any additional variables to write this IF condition; all the information we need is in the data we have. (BTW, please use the desc variable to distinguish between notices of sale and sales. Some sales are missing sales price, for instance.) 

Once you have written that IF statement, we need to think about what happens if that condition isn't true. That is, we need to check whether it meets the other condition for u_dedup_notice to be set to 1: it is a NOTICE OF SALE that was preceded by a SALE for the same u_address_id_ref. 

Now, if you are looking at a NOTICE OF SALE observation, we don't have anything currently that would tell us that it was preceded by a SALE. So we need to create a variable with that information. **Think about how you would create that variable using RETAIN.** 

I also want to mention that all this can be done in a single data step. 

Let's focus on getting this far first and then we can build on that. If it would be helpful to talk it through, let's find some time to do that. Thanks!

ptatian:
Hi @eburton9. Thanks for the first pass at this. Data step programming can be tricky stuff, so let's take it one piece at a time. Starting with the u_dedup_notice variable. It isn't quite right yet. I added a Proc Print to the program that will show what things look like for the example I included above, so that we can see how that compares to our current code. Here's what it looks like after the Topa_notice_flag data step. 

**DC MAR address ID=5142**

ref_date | desc | ID | temp1 | u_dedup_notice 
-- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 
05/13/2014 | SALE | . | 1 | 0 
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 
11/07/2006 | SALE | . | 1 | 0 
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 

We want u_dedup_notice = 1 for notice IDs 1449, 764, and 37, and = 0 otherwise. 1449 because it is the first obs (in this sort order) for the u_address_id_ref AND it is a NOTICE OF SALE. 764 and 37 because they are the last notices before a SALE (that is, they immediately follow a SALE in this sort order). 

The temp1 flag you have now is always 1 in this case, so not especially useful. It looks like you may be trying to use it to flag the first obs for the property, but it isn't necessary to create a new var for that because that is what first.u_address_id_ref tells you. Think of first.u_address_id_ref like another var, but one that isn't saved to the data set. If you were to actually see it within the data step, here's what it would look like. 

ref_date | desc | ID | temp1 | u_dedup_notice | first.u_address_id_ref
-- | -- | -- | -- | -- | --
05/20/2020 | NOTICE OF SALE | 1449 | 1 | 1 | 1
05/13/2014 | SALE | . | 1 | 0 | 0
03/06/2014 | NOTICE OF SALE | 764 | 1 | 0 | 0
08/01/2013 | NOTICE OF SALE | 667 | 1 | 0 | 0
11/07/2006 | SALE | . | 1 | 0 | 0
10/06/2006 | NOTICE OF SALE | 37 | 1 | 0 | 0

Back to u_dedup_notice. **First, we want to write an IF statement to set u_dedup_notice to 1 if the observation is the first one for u_address_id_ref and it is a notice of sale.** You do not need to create any additional variables to write this IF condition; all the information we need is in the data we have. (BTW, please use the desc variable to distinguish between notices of sale and sales. Some sales are missing sales price, for instance.) 

Once you have written that IF statement, we need to think about what happens if that condition isn't true. That is, we need to check whether it meets the other condition for u_dedup_notice to be set to 1: it is a NOTICE OF SALE that was preceded by a SALE for the same u_address_id_ref. 

Now, if you are looking at a NOTICE OF SALE observation, we don't have anything currently that would tell us that it was preceded by a SALE. So we need to create a variable with that information. **Think about how you would create that variable using RETAIN.** 

I also want to mention that all this can be done in a single data step. 

Let's focus on getting this far first and then we can build on that. If it would be helpful to talk it through, let's find some time to do that. Thanks!

eburton9:
@ptatian I think I now actually have both u_dedup_notice and u_notice_with_sale working, thank you for meeting with me this morning. I'll wait for your review on the flags and then work on u_days_from_dedup_notice_to_sale and retaining the sale information on the notice observations. Let me know if I'm missing any next steps. Thanks!

ptatian:
Thanks, @eburton9. Unfortunately, SAS1 is down again for me. But what you wrote looks good so when don't you carry on with the next steps.

eburton9:
@ptatian I calculated days_from_dedup_notice_to_sale and days_between_notices in the data step. I also retained the sale information in [data analysis plan 5.a](https://docs.google.com/document/d/1hnet3wDgDR876SYBmAefSgXPnmcH6Ybyqj7bJ3xrHoo/edit?usp=sharing), which included keeping the buyer address vars earlier in the prog. However, I don't see a variable in the four TOPA databases for #4 in the analysis, "Buyer type (individual, for profit co./LLC, nonprofit, other)." Also, I'm unsure how to take out the desc="SALE" without the rest of the newly created vars disappearing (tried a different data step as well but it seems like the new vars from retain functions don't keep?). Lastly, retaining the sale info keeps it for all the notices not just the one where dedup=1, and my attempts at clearing those out didn't work either. I'm assuming this doesn't work because it affects the data steps I just did. Thanks!

ptatian:
Thanks @eburton9 . Looks great! Yes, there is a problem with the SALES data being carried forward. The reason you couldn't solve it was because we need to introduce an OUTPUT statement to control when the data are saved to the new data set. Resetting the sales info has to happen _after_ we write the notice observation. 

```
  if desc="SALE" then do; 
	u_ownername=Ownername_full; u_saleprice=SALEPRICE; u_proptype=ui_proptype; u_address1=ADDRESS1;
	u_address2=ADDRESS2; u_address3=address3; 
	end; 
  retain u_ownername u_saleprice u_proptype u_address1 u_address2 u_address3; 

  /** Write observation if a notice of sale and reset retained sales data for next observation **/
  if desc="NOTICE OF SALE" then do;
	output;
	u_ownername=""; u_saleprice=.; u_proptype=.; u_address1="";
	u_address2=""; u_address3=""; 
	end;
  ```

The default for a data step is to save the data to the data set at the very end. By including an OUPUT statement, we change that so that the data get saved at that point. Note that by including it in the IF statment, only NOTICE OF SALE obs will be saved.  I've added a temporary second OUTPUT statement so that we can see the sales records, too, to help with checking. 

```
  /** TEMPORARY CODE - REMOVE BEFORE FINALIZING PROGRAM **/
  IF DESC="SALE" THEN OUTPUT;
```

But we will remove the above before we finalize the program. 

I also changed the PROC PRINT so that a few more address_id's are being shown and we can check how things are working when going from one address_id to another. The sales data was also being carried over to other addresses before, but the above code fixes that.

eburton9:
@ptatian It looked like sale_date was staying from preceding sale obs from different address_ids so I added u_sale_date=.; in your output statement. Also updated/added the three things to finish up you listed above. It looks good to me!

ptatian (approved):
Hi @eburton9 Looks good to me, too! Thank you for catching the problem with u_saleprice. You can go ahead and merge and do final submit on SAS1. 
I made a few tweaks and, for your awareness, I'll put some notes in the conversation to explain what I did. But we can now move on to getting the spreadsheet that Martha wants. Thanks!

ptatian:
Just a few notes on my changes:

- For 0/1 vars, we have a dyesno. format that we like to use. This will show up as Yes/No in the metadata.
- I added "(Urban created var)" to the var labels. It's tedious but later it will be necessary to document what data we added vs what data came from CNHED. 
- I also changed some of the owner and sale var labels to make it clear that this was information about the property buyer, and not the seller.
- Added more explanation to the labels for u_address_id_ref and fulladdress.
- Dropped the desc var from the final data set, since it is not needed. (Everything in the data set is a notice.) 
- Kept the Proc Print at the end but included fewer vars so that the output is easier to read.
- Removed the last %File_info() reference because %Finalize_data_set() does the same thing. 
- In %Finalize_data_set(), removed "data" from the name of the final data set. (Not needed, since it is a data set.) Included the notice year range in the data set label. Added a _freqvars=_ parameter to produce frequency tables for a few key vars.

PR #356, opened 2023-03-30
Create TOPA spreadsheet for CBO TA providers

eburton9:
@ptatian Ready for your review.. I manually created vars and sorted them (I'm sure there's a more sophisticated way to do that?). One question, the label names for our variables are different than on the mock spreadsheet, do those labels needed to be edited to exactly match the mock spreadsheet? Oh also, does the spreadsheet need both the offer of sale date and the CASD date?

### DIFF
--- a/Prog/TOPA/TOPA_CBO_spreadsheet.sas
+++ b/Prog/TOPA/TOPA_CBO_spreadsheet.sas
@@ -87,7 +87,7 @@ options missing=' ';
 
 ** Export for CBO spreadsheet **;
 ods tagsets.excelxp   /** Open the excelxp destination **/
-  file="&_dcdata_default_path\PresCat\Prog\TOPA\Topa_CBO_sheet.xls"  /** This is where the output will go **/
+  file="&_dcdata_default_path\PresCat\Prog\TOPA\Topa_CBO_sheet_newsales.xls"  /** This is where the output will go **/
   style=Normal    /** This is the ODS style that will be used in the workbook **/
   options(sheet_interval='Proc' )
 ;
@@ -96,7 +96,7 @@ ods listing close;  /** Close the regular listing destination **/
 ods tagsets.excelxp options( sheet_name="With Sales" );
 
 proc print label data=Topa_CBO_sheet_retain;
-  where u_dedup_notice=1 and u_notice_with_sale=1;
+  where u_dedup_notice=1 and u_notice_with_sale=1 and u_sale_date < '01jan2021'd;
   id id; 
   var u_address_id_ref u_notice_date All_street_addresses Property_name u_date_dhcd_received_ta_reg u_sale_date u_ownername r_notes r_TA_provider
 r_TA_staff r_TA_lawyer r_TA_claim_rights r_TA_dev_partner TA_negotiate r_date_TA_ass_rigts r_dev_ass_right ass_aff_developer dev_agree buyouts assign_terms add_notes;
@@ -106,6 +106,15 @@ ods tagsets.excelxp options( sheet_name="Without Sales" );
 
 proc print label data=Topa_CBO_sheet_retain;
   where u_dedup_notice=1 and u_notice_with_sale=0;
+  id id; 
+  var u_address_id_ref u_notice_date All_street_addresses Property_name u_date_dhcd_received_ta_reg u_sale_date u_ownername r_notes r_TA_provider
+r_TA_staff r_TA_lawyer r_TA_claim_rights r_TA_dev_partner TA_negotiate r_date_TA_ass_rigts r_dev_ass_right ass_aff_developer dev_agree buyouts assign_terms add_notes;
+run;
+
+ods tagsets.excelxp options( sheet_name="Sales in 2021 and 2022" );
+
+proc print label data=Topa_CBO_sheet_retain;
+  where u_dedup_notice=1 and u_notice_with_sale=1 and u_sale_date > '31dec2020'd;
   id id; 
   var u_address_id_ref u_notice_date All_street_addresses Property_name u_date_dhcd_received_ta_reg u_sale_date u_ownername r_notes r_TA_provider
 r_TA_staff r_TA_lawyer r_TA_claim_rights r_TA_dev_partner TA_negotiate r_date_TA_ass_rigts r_dev_ass_right ass_aff_developer dev_agree buyouts assign_terms add_notes;
--- a/Prog/TOPA/TOPA_CBO_spreadsheet.sas
+++ b/Prog/TOPA/TOPA_CBO_spreadsheet.sas
@@ -23,7 +23,7 @@
 /** Fill in missing columns from original TOPA database to notices sale database created**/
 data Topa_CBO_sheet; 
   merge 
-    PresCat.TOPA_notices_sales (keep=id u_address_id_ref u_dedup_notice u_notice_date u_ownername u_sale_date u_notice_with_sale in=time_period)
+    PresCat.TOPA_notices_sales (keep=id u_address_id_ref u_dedup_notice u_notice_date u_ownername u_sale_date u_notice_with_sale u_sum_units u_year_built_original u_recent_reno Ward2022 in=time_period)
     Prescat.Topa_database (keep=id All_street_addresses Property_name Notes Technical_assistance_provider u_date_dhcd_received_ta_reg Tech_Assist_Staff Tenant_Assn_Lawyer TA_Development_Partner Existing_LIHTC_Financing New_LIHTC_Financing);
   by id;
   if time_period;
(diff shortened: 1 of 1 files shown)


Prog/TOPA/TOPA_CBO_spreadsheet.sas
```diff
@@ -0,0 +1,58 @@
...
+%DCData_lib( PresCat )
+
+%File_info( data=PresCat.TOPA_database, printobs=5 ) 
+%File_info( data=PresCat.TOPA_notices_sales, printobs=5 ) 
+
+/** Fill in missing columns from original TOPA database to notices sale database created**/
+data Topa_CBO_sheet; 
+  merge 
+    PresCat.TOPA_notices_sales (keep=id u_address_id_ref u_dedup_notice u_notice_date u_ownername u_sale_date in=time_period)
+    Prescat.Topa_database (keep=id All_street_addresses Property_name Date_DHCD_received_TA_reg Notes Technical_assistance_provider Tech_Assist_Staff Tenant_Assn_Lawyer Did_a_TA_claim_TOPA_rights TA_Development_Partner Date_TA_assignment_of_rights Developer_assignment_of_rights);
```
ptatian, line 27: @eburton9 I added a new variable, u_Date_DHCD_received_TA_reg, to TOPA_database. You should use that instead of Date_DHCD_received_TA_reg.

Prog/TOPA/TOPA_CBO_spreadsheet.sas
```diff
@@ -0,0 +1,58 @@
...
+  if time_period;
+  TA_negotiate=""; label TA_negotiate = "Did TA negotiate outside of TOPA process?";
+  ass_aff_developer=""; label ass_aff_developer ="Is Assignee an Affordable Housing developer?";
+  dev_agree=""; label dev_agree = "Development agreement? (Y/N)";
+  buyouts=""; label buyouts = "Number of buy outs"; 
+  assign_terms=""; label assign_terms = "Description of assignment terms";
+  add_notes=""; label add_notes = "Additional notes"; 
+run; 
+
+%File_info( data=Topa_CBO_sheet, printobs=5 )
```
ptatian, line 38: @eburton9 We want the data in the spreadsheet to be sorted by u_address_id_ref and u_notice_date, so please add a Proc Sort here.

Prog/TOPA/TOPA_CBO_spreadsheet.sas
```diff
@@ -0,0 +1,58 @@
...
+** Export for CBO spreadsheet **;
+ods tagsets.excelxp   /** Open the excelxp destination **/
+  file="&_dcdata_default_path\PresCat\Prog\TOPA\TOPA_CBO_TA.xls"  /** This is where the output will go **/
+  style=Normal    /** This is the ODS style that will be used in the workbook **/
+;
+
+ods listing close;  /** Close the regular listing destination **/
+proc print label data=Topa_CBO_sheet;
+  where u_dedup_notice=1;
+  var u_address_id_ref id u_notice_date All_street_addresses Property_name Date_DHCD_received_TA_reg u_sale_date u_ownername Notes
```
ptatian, line 50: @eburton9 Remove ID variable from the VAR statment and add it to a new ID statement. This will make the ID variable be the far left column that identifies each row.

ptatian:
@eburton9 This is super! A few small tweaks above. 

My only potential concern is whether any of the technical assistance data that we are including (Technical_assistance_provider, Tech_Assist_Staff, etc.) might have been entered on one of the earlier notice records and not in the dedup notice. Would you please take a look at the data and see if that is something we need to worry about?

eburton9:
@ptatian Edited small tweaks but when looking at first 20 obs I see one instance where TA staff, provider, lawyer, and TA claiming TOPA rights was in the earlier notice record and not in the dedup notice.. I can work on using retain to do that tomorrow morning.. apologies didn't have time this afternoon.

Prog/TOPA/TOPA_CBO_spreadsheet.sas
```diff
@@ -35,19 +35,23 @@ data Topa_CBO_sheet;
   add_notes=""; label add_notes = "Additional notes"; 
 run; 
 
-%File_info( data=Topa_CBO_sheet, printobs=5 ) 
+proc sort data=Topa_CBO_sheet out=Topa_CBO_sheet_sorted;
+  by u_address_id_ref descending u_notice_date;
```
ptatian, line 39: Not descending this time. Just normal order.

ptatian:
Thanks, @eburton9. That's what I was worried about. Thanks for taking a look at it! Let's see if we can wrap it up tomorrow!

Prog/TOPA/TOPA_CBO_spreadsheet.sas
eburton9: @ptatian Yep - I'll change it back i just did that to look at the dedup_notices that way.

eburton9:
@ptatian Retaining the dedup="no" vars needed, but I don't want to drop the other variables from dedup=yes.. should these be renamed or a way to concat info?

ptatian:
@eburton9 I think what you have is good. The missing part is that we want to use data that is in a dedup=yes obs by default, and only use the dedup=no information if the dedup=yes info is missing. So adding something like this to your IF block before the OUTPUT statement:

```
if not( missing( Notes ) ) then r_notes=Notes;
if not( missing( Technical_assistance_provider ) ) then r_TA_provider = Technical_assistance_provider;
etc.
```

Actually, you could use the same code in the earlier dedup=no assignments, so that blank data does not overwrite actual data. 

Let me know how that sounds.

eburton9:
@ptatian Okay the output excel sheet looks good to me! Ready for your review.

Prog/TOPA/TOPA_CBO_spreadsheet.sas
```diff
@@ -0,0 +1,104 @@
...
+  label r_TA_staff='Technical assistance staff';
+  label r_TA_lawyer='Technical assistance lawyer';
+  label r_TA_claim_rights='Did TA provider claim TOPA rights?';
+  label r_TA_dev_partner='TA Development Partner';
+  label r_date_TA_ass_rigts='Approx. Date of TA assignment of rights';
+  label r_dev_ass_right='Developer receiving assignment of rights';
+
+/*  if u_dedup_notice=1 then do;*/
+  if u_dedup_notice=1 and not( missing( Notes ) ) then r_notes=Notes;
+  else if u_dedup_notice=1 not( missing( Technical_assistance_provider ) ) then r_TA_provider=Technical_assistance_provider;
```
ptatian, line 71: You don't want to use ELSE IF here because the conditions are not mutually exclusive. That is, you can have missing values for Notes but not for Technical_assistance_provider. So they should all be independent IFs. 

Second, you left out the AND between the two conditions. `u_dedup_notice=1 AND not( missing( ...`  But, also, it is better to use the DO; ... END; block structure that you had earlier. If you are writing the same condition repeatedly (u_dedup_notice=1) then it is usually better to make that an outer IF and then have the other statements within a DO; ... END; block. That's the same as writing AND... repeatedly, but it is easier to read and more efficient.

ptatian: Some additional background on IF THEN ELSE. http://course1.winona.edu/thooks/Media/Handout%204%20-%20If-Then%20Statements%20in%20SAS.pdf

eburton9:
@ptatian Whoops you're right I didn't save my newest version but I fixed the if statements now too. Double checked and now things should be working correctly.

ptatian (approved):

ptatian:
@eburton9 Looks great! They may want changes so let's leave this open for now--don't merge or close.

ptatian:
Sigh... Now noticing that we have sales information in the rows that are supposed to be without sales. These are all sales that happened before the notice, so it seems the notice with sale flag is working properly and we have the right notices in the respective worksheets. But it might confuse people to see the sales info in the notices without sales. 

Perhaps just manually delete the sales information in the Without Sales sheet for now and we can go back and fix the other code later if we need to.

eburton9:
@ptatian Made edits to the TOPA CBO spreadsheet program. It looks good on my end. One thing I noticed is there's 45 less notices without sales in this updated spreadsheet than the old one but I didn't change anything in that code..

ptatian:
Hmm...I assume that means there are 45 more notices with sales? Can you identify any of the notices that switched category?

eburton9:
@ptatian Yes - it looks like the 45 less notices without sales are in the 2021-2022 sales tab - but do we want those duplicated in the without sales tab as well?

ptatian:
@eburton9 Oh, wait. **Of course** there are fewer no sale notices. Because we have 45 notices with sales in 2021-2022 that we weren't counting as having a sale previously. So this all seems fine.

ptatian:
Would you please go ahead and send this to Martha. Thanks!

eburton9:
@ptatian Please let me know if this is okay and I'll save as Excel and make columns less wide.

ptatian:
Hi @eburton9 This looks great. Please see the email I just sent to everyone on next steps. We do need to tweak the formatting a bit because Ryan and Martha made changes to a few of the columns.

eburton9:
@ptatian Realized I had the LIHTC columns in the wrong place

ptatian:
@eburton9 This looks good. Let's check with Martha before updating the Google Sheet.

eburton9:
@ptatian I added the 4 variables Martha wanted to the CBO spreadsheet tabs. I just put them at the end columns but let me know if they should be elsewhere in the spreadsheet. I'll save shet as xlsx and reformat columns once I get your ok.

ptatian (approved):
@eburton9 Good to go!

Issue #357, opened 2023-04-05
adding 2020-2022 sale data


eburton9:
updating TOPA_match_notices_property

PR #358, opened 2023-04-05
Update #357

eburton9:
@ptatian Created two new data steps to filter the years for the notices and for the sale dates on the prop data.. probably is a more sophisticated way to do that but adding the where statement to the earlier data steps with merges doesn't work..

### DIFF
--- a/Prog/TOPA/TOPA_match_notices_property.sas
+++ b/Prog/TOPA/TOPA_match_notices_property.sas
@@ -95,7 +95,7 @@ run;
 
 data Combo;
   set 
-    TOPA_by_property 
+    TOPA_by_property_dates 
     (keep=u_address_id_ref id u_offer_sale_date FULLADDRESS Anc2012 Geo2020 GeoBg2020 GeoBlk2020 Psa2012 VoterPre2012 Ward2012 Ward2022 cluster2017
      rename=(u_offer_sale_date=u_ref_date)
      in=is_notice)
@@ -103,21 +103,19 @@ data Combo;
 	  (keep=u_address_id_ref saledate saleprice ownername_full ui_proptype ADDRESS1 ADDRESS2 address3
 	   rename=(saledate=u_ref_date));
 	by u_address_id_ref descending u_ref_date;
-
 	length desc $ 40;
 
 	if is_notice then desc = "NOTICE OF SALE";
 	else desc = "SALE";
 
-    /**Limit sale and notice data to 2006-2020 and do not use obs with missing address or date **/
-    where u_ref_date between '01Jan2006'd and '31dec2020'd 
-	  and not( missing( u_address_id_ref ) or missing( u_ref_date ) );
+    /**do not use obs with missing address or date **/
+    where not( missing( u_address_id_ref ) or missing( u_ref_date ) );
 
 run;
 
 %File_info( data=Combo, printobs=5 ) /** 4050 obs**/
 
-/*Create flags (u_dedup_notice & u_notice_with_sale), u_days_between_notices & u_days_from_dedup_notice_to_sale*/
+/*Create flags (u_dedup_notice & u_notice_with_sale & u_sale_after_2020), u_days_between_notices & u_days_from_dedup_notice_to_sale*/
 data Topa_notice_flag; 
   set Combo;  
   by u_address_id_ref descending u_ref_date;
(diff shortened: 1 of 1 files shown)


ptatian:
@eburton9 Yes, the way you went about it makes sense!

Let's also investigate the issue around why u_notice_with_sale=0 but there is sales information.  Earlier, it looked like these were cases where the sale date was prior to the notice date. But in this new version, I'm seeing some cases where the sale date is after the notice date. It looks like the second group is sales in 2021-2022, but I didn't check thoroughly. 

Also, small point, but you should update the revisions=%str(New data set.) message to describe briefly the changes we are making to the data set in this program.

eburton9:
@ptatian Could you point out some instances? I'm not seeing that but maybe I'm not understanding. What do you mean by second group?

ptatian:
```
Unique property ID (DC MAR address ID) (Urban created var)=30765
                                                                     u_days_from_     u_days_
                u_notice_       u_sale_    u_dedup_    u_notice_    dedup_notice_    between_
 Obs     ID          date          date     notice     with_sale       to_sale        notices    u_saleprice

 202    182    02/20/2008    02/03/2022      Yes          No              .              .            0     
 
 Obs         u_ownername                        u_proptype

 202    NEW ENDEAVORS BY WOMAN    Residential: Rental apartment building
```

eburton9:
Looking at u_address_id_ref = 58024 and u_address_id_ref = 30765 in the Combo dataset and compared to topa_flag_notice dataset. For 58024 it looks likeit's pulling the sale info from the sale preceding it that is a different u_address_id_ref (which also doesn't even have a notice with the same u_address_id_ref so maybe could help by not including those..). In my recent commit I put two comments where I thought may there should be an additional conditional statement to make sure the u_address_id_ref match but I don't know how to go about that. I also tried adding the first.u_address_id to be blank at the beginning of the data step like with the u_notice_date but that didn't work.

ptatian:
Thanks @eburton9. What I would recommend is setting all retained variables to missing values (. or "") if first.u_address_id_ref is true. That would eliminate any carry-over from previous properties.

eburton9:
@ptatian Yes, I did that yesterday (but just cleaned it up in the previous commited), it helps with the dedup notice and notice with a sale flag but for example when doing  proc print data=Topa_notice_flag; where u_address_id_ref = 58024; the sale info from a separate u_address_id (comparing to Combo set) is showing up as a 58024 sale even though it's not.. the obs doesn't have any of the flags - so it wouldn't show up in our output that we sent to Martha (and maybe it isn't important?) but could be happening with other instances where it does affect that.

ptatian:
@eburton9 Look at the order of your output statement and the statements setting the retained data to missing. You are writing the data _before_ you set the values to missing.

ptatian:
I'm not sure why you have an output statement there, however. Is that just for testing?

eburton9:
@ptatian Yeah i was realizing I wasn't understanding the output statement at all... but i read about it and i don't need it there i don't think

ptatian:
Looks better! Let me check it out.

ptatian:
@eburton9 The new initialization of the sales variables seems to have fixed the flags, but it created another problem. Can you see what it is?

eburton9:
@ptatian I updated the format of the owner name so it wasn't getting cut off. I compared the outputs and I think that was the only issue.

ptatian:
Yes, that was it. Well spotted! Because we added the initialization of the owner name variable, and that was the first reference to that var in the data step, that automatically set the length of that variable to only what was needed for the initialization value (blank). So later data was getting truncated. 

By adding the FORMAT statement you signaled to SAS that this character variable needed to be longer. A LENGTH statement would have done the same thing. LENGTH sets the variable size without adding the format. 

This is something to watch for with character variables. Numeric variables have a default length that is at the maximum.

eburton9:
That makes sense! Thank you. Are you still reviewing this?

ptatian:
@eburton9 Yes. Actually, the same problem needs to be fixed for the u_address... variables.

eburton9:
@ptatian whoops was only looking at the proc print. it's good now

ptatian:
@eburton9 I missed it on the first go around, too. Looks good now! You can finalize this.

ptatian (approved):

Issue #345, opened 2023-03-16
Make edits to Prescat.Subsidy to adjust LIHTC subsidy records


ptatian:
Need to make final fixes to the 2020 LIHTC update. First changes to **Prescast.Subsidy**:
- Add new tax credit record for Cedar Heights to Finsbury Square Apts (NL000106). 
- Add new tax credit record to Portner Flats (NL000243).
- Remove new tax credit record from Augusta Louisa (NL000337). 

One change to Prescat.Project_category
- Rename Finsbury Square Apts to Cedar Heights.

Instructions for these steps are below. 

***
### Changes to Prescat.Subsidy

This task involves three modifications to **Prescat.Subsidy**. We'll write one new SAS program to do all this, and all the steps I describe below will go into this new program. 

In **Prescat.Subsidy**, there is one row for each subsidy associated with a project. Here's a (made up) example. 

| Nlihc_id | Subsidy_id | Portfolio | Subsidy_source_info_date | _More variables_ |
| --- | --- | --- | --- | --- |
| NL000001 | 1 | LIHTC | 4/2/2018 | ... |
| NL000001 | 2 | CDBG| 5/7/2019 | ... |
| NL000002 | 1 | PUBHSG | . | ... |
| _and so on_ |

Each row is uniquely identified by Nlihc_id and Subsidy_id. Subsidy_id just numbers the subsidy records 1, 2, 3, ..., depending on how many subsidies are associated with the project. So to add the new subsidy rows we need to first figure out which number they should be. You can use Proc Print to display the subsidy records for the two projects we are interested in. 

```
proc print data=Prescat.Subsidy;
  where nlihc_id in ( 'NL000106', 'NL000243' );  /** Limit output to rows with these project IDs **/
  var nlihc_id subsidy_id program Subsidy_info_source Subsidy_source_info_date;  /** Limit output to these variables **/
run;
```

In the output, you'll see the listing of the subsidy rows, which tells us what the number should be of the next row for the project. For instance, if NL000106 currently has 3 rows, the next would by subsidy_id = 4. 

[...]

PR #359, opened 2023-04-18
Update #345

dharvey17:
Ready for your to look @ptatian

### DIFF
--- a/Prog/Dev/345_edit_LIHTC_records.sas
+++ b/Prog/Dev/345_edit_LIHTC_records.sas
@@ -37,23 +37,33 @@ data Subsidy_new_obs;
 
   /** Data that is the same for both projects **/
 
-  Subsidy_source_info_date = ;
+  Subsidy_source_info_date = 4/8/2022;
 
-  Subsidy_source = "";
+  Subsidy_source = "HUD/Low Income Housing Tax Credits";
 
-  /** New tax credit row for Cedar Heights/Finsbury Square Apts (NL000106) **/
+  /** New tax credit row for Cedar Heights/Finsbury Square Apts (NL000106)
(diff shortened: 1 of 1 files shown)


ptatian:
Hi @dharvey17 I figured out that the SET statement with OBS=0 was the problem. So I just took that out. The only thing that adds is that we should declare lengths of any character variables first using a LENGTH statement. I added a couple of examples and am asking you to fill in the remaining ones. 

Also, `= 4/8/2022` is not the correct way to specify a date value in SAS. (It actually just divides 4 by 8 by 2022.) The two options are: 
- Use the mdy() function: `= mdy( 4, 8, 2022 )`
- Use a SAS literal date value: `= '8apr2022'd`

ptatian:
On Cedar Heights, maybe I got confused by all the name changes, etc. It seems like the contract that was in the 2020 update is already in Prescat.Subsidy, so there really isn't anything to add. We can skip that step.

ptatian:
But we still want to do the project rename from Finsbury Square Apts to Cedar Heights.

dharvey17:
@ptatian ready for review.

ptatian:
Hi @dharvey17 For your awareness, I am reviewing this, but found a problem with one of the previous updates that I'm investigating further. Hope to get back to you shortly. Thank you for your patience.

ptatian (approved):
@dharvey17 This is good to go. I had to fix some things with variable names that might have been my fault. I also seemed to have still been confused about Augusta Louisa, but got it sorted out. Felt it was easier to just make those corrections directly. I added a couple of Proc Compare's, which I sometimes use to make sure that changes I'm introducing to files are only what I expect. 

You can go ahead and finalize this. Thanks!

PR #363, opened 2023-04-27
create basic descriptive tables

eburton9:
@ptatian Create the 6 tables, ready for your review. Interesting to see! Not surprised that the neighborhood cluster I live in has the most units and deduplicated notices. Also interesting that  Ward 1 had higher notices earlier and Ward 8 later and the obvious gentrification patterns.

### DIFF
--- a/Prog/TOPA/TOPA_desc_tables.sas
+++ b/Prog/TOPA/TOPA_desc_tables.sas
@@ -49,7 +49,7 @@ run;
 
 ** Printing Descriptive Tables **;
 options nodate nonumber;
-options orientation=portrait;
+options orientation=landscape;
 
 %fdate()
 
--- a/Prog/TOPA/TOPA_desc_tables.sas
+++ b/Prog/TOPA/TOPA_desc_tables.sas
@@ -47,18 +47,19 @@ run;
 
 %File_info( data=TOPA_table_data)
 
-** 1a. Table All Notices by Ward and Year **;
+** Printing Descriptive Tables **;
 options nodate nonumber;
 options orientation=portrait;
 
 %fdate()
 
 ods listing close;
-ods rtf file="&_dcdata_default_path\Prescat\Prog\Topa\TOPA_desc_tables.rtf" style=Styles.Rtf_lato_9pt startpage=proc;
+ods rtf file="&_dcdata_default_path\Prescat\Prog\Topa\TOPA_desc_tables.rtf" style=Styles.Rtf_lato_9pt;
 
 footnote1 height=9pt "Prepared by Urban-Greater DC (greaterdc.urban.org), &fdate..";
 footnote2 height=9pt j=r '{Page}\~{\field{\*\fldinst{\pard\b\i0\chcbpat8\qc\f1\fs19\cf1{PAGE }\cf0\chcbpat0}}}';
 
+** 1a. Table All Notices by Ward and Year **;
 proc tabulate data=TOPA_table_data /** insert name of input data set here **/ format=comma12.0 noseps missing;
   class ward2022 u_notice_date;   /** These variables define the table rows and columns **/
   var all_notices;  /** This variable is used for the content of the table **/
@@ -80,11 +81,6 @@ proc tabulate data=TOPA_table_data /** insert name of input data set here **/ fo
   title3 "1a. TOPA Notices of Sale (With Duplicates) by Ward and Year, 2006-2020";
 run;
 
-title2;
-footnote1;
-
-ods rtf close;
-ods listing;
 
 ** 1b. Table All Notices by Neighborhood Cluster and Year **;
 proc tabulate data=TOPA_table_data format=comma12.0 noseps missing;
@@ -190,3 +186,9 @@ proc tabulate data=TOPA_table_data format=comma12.0 noseps missing;
   title2 " ";
   title3 "3b. Residential Units in Properties with TOPA Notices of Sale by Neighborhood Cluster and Year, 2006-2020";
 run;
+
+title2;
+footnote1;
+
+ods rtf close;
+ods listing;
--- a/Prog/TOPA/TOPA_desc_tables.sas
+++ b/Prog/TOPA/TOPA_desc_tables.sas
@@ -183,6 +183,7 @@ run;
 
 ** 3a. Table Residential Units by Ward and Year**;
 proc tabulate data=TOPA_table_data format=comma12.0 noseps missing;
+  where u_dedup_notice=1; 
   class ward2022 u_notice_date;   
   var u_sum_units;  
   table 
@@ -199,11 +200,12 @@ proc tabulate data=TOPA_table_data format=comma12.0 noseps missing;
   ;
   format u_notice_date year.;  
   title2 " ";
-  title3 "3a. Residential Units in Properties with TOPA Notices of Sale by Ward and Year, 2006-2020";
+  title3 "3a. Residential Units in Properties with TOPA Notices of Sale (Deduplicated) by Ward and Year, 2006-2020";
 run;
 
 ** 3b. Table Residential Units by Neighborhood Cluster and Year**;
 proc tabulate data=TOPA_table_data format=comma12.0 noseps missing;
+  where u_dedup_notice=1; 
   class cluster2017 u_notice_date;   
   var u_sum_units ;  
   table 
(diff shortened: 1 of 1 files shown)


ptatian:
Hi @eburton9 This looks good! Exciting to see some results. 

I see what you did with the TOPA_sum_units Data step. It's generally better to use a Proc rather than a Data step if you can. In this case, [you could use Proc Summary](https://www.lexjansen.com/pharmasug/2009/tt/TT05.pdf) to summarize the data by ID. [Proc Means and Proc Summary do the same things, but the defaults are different. So you can use either. It's a bit of weird historical quirk in SAS.]

ptatian:
Next step is to make the output nicer looking. We'll do that by using SAS ODS to create Word compatible tables. There is some code to put around your Proc Tabulate code. 

```
options nodate nonumber;
options orientation=portrait;

%fdate()

ods listing close;
ods rtf file="&_dcdata_default_path\Prescat\Prog\Topa\TOPA_desc_tables.rtf" style=Styles.Rtf_lato_9pt startpage=proc;

footnote1 height=9pt "Prepared by Urban-Greater DC (greaterdc.urban.org), &fdate..";
footnote2 height=9pt j=r '{Page}\~{\field{\*\fldinst{\pard\b\i0\chcbpat8\qc\f1\fs19\cf1{PAGE }\cf0\chcbpat0}}}';


{PROC TABULATE CODE HERE}


run;

title2;
footnote1;

ods rtf close;
ods listing;
```

ptatian:
For the above to work properly, you need to submit this program `\\sas1\DCData\Libraries\General\Prog\Style_rtf_lato_9pt.sas` to create the new ODS style. 

You can either open the program in the SAS for Windows app and submit it from the program editor window, or you can batch submit it in Windows File Explorer. You should only need to do this one time and it will permanently save the new style to your SAS set up. (This does not require opening a new issue or branch.)

ptatian:
@eburton9 Yes, seems like I got that wrong. Just take out the startpage=proc entirely. I think it is OK without it.

ptatian:
I'll take a look.

eburton9:
I just changed the orientation to landscape

ptatian:
This looks good! Maybe we can also do a table showing deduplicated notices that resulted in a property sale. This could be a separate number (count) table and a percentage table. To do a percentage, you can change the statistic in Proc Tabulate from SUM to MEAN and take the average of the 0/1 variable. You also want to assign the percent format to the 0/1 variable on the Proc Tabulate FORMAT statement. 

In the above, use a WHERE statement to limit the notices to u_dedup_notice=1. And then use the u_notice_with_sale variable as the analysis VAR, which will get SUM'd or MEAN'd to get the count and percentage statistics, respectively.

eburton9:
@ptatian Added 4 more tables for notices with sales but my mean statistic isn't working for the percentage ones.. not sure if how I have it is showing up as 1s because it's only averaging if notice_with_sale = 1 instead of all the notices? I looked through resources and I couldn't find a quick fix but I know you're wanting to show tables to CNHED and Martha tomorrow so wanted to tag you ASAP

eburton9:
@ptatian I added tables 6a and 6b that are Residential Units in Properties with TOPA Notices that Sold. I also compared the u_sum_units var I created in Topa_notices_sale dataset to the Units var in the TOPA_database and it seems like most of them are the same (sometimes 1-3 units off, more units in u_sum_units then when Units says 5+, or Units is blank). Looking at Ward 6 specifically, IDs 316, 336,  753, 754, 862, 1260 all have 100+ more units in the u_sum_units var than the Units var... may be worth looking into those? But besides that they were mostly the same. I can look more closely at other wards too.

ptatian:
Thanks, @eburton9. Those IDs certainly sound like they are worth looking at more closely. Can you check the Topa_addresses data set for those IDs and see what that looks like? From the intern notes, 316 looks like it was a condo conversion and I'm wondering if that may be causing us to overcount units. It might be also helpful to check a few properties manually in [the MAR Address Lookup](https://developers.data.dc.gov/marviewer/home). If you click on the Units tab in the results, it will show you how many units MAR has for the property.

eburton9:
@ptatian Looking at TOPA_addresses IDs: 
1. 316 MAR address says our u_sum_unit is wrong TOPA_database is right. Under units in MAR it says no condo but under location and SSL it says condo... 
2. 336 MAR address says 12 units not 29 (which our u_sum_unit says (no condos)
3. 753 and 754 seem like they're the same property (so duplicating counting) because they have the same address_id and mapping the addresses in TOPA_database they look like they're the same apartment building. But looking up manually in MAR - our number of units is correct (no condos) 
4. 862 Condo building that we are overcounting by 636 units...
5. 289 ID also overcounting by ~300 units.. but it's not a condo. 

For all the IDs were overcounting it seems like in the TOPA_addresses database there is one address_id that has the correct number of units but that we're counting units from address_ids that shouldn't be connected to the address but seem to be nearby. So maybe it's an issue with how we're joining those..

ptatian:
Thanks, @eburton9. Yes, it's complicated, but I actually think what we are doing may be right. For ID=316, for instance, there's a single record lot, [0515 0158](https://developers.data.dc.gov/marviewer/search/ssl/find?square=0515&lot=0158), that ties all these addresses together. They are separate buildings, but it was a single piece of real property until it was divided up into condominiums. 

Same with ID=862, which has record lot [0542 0089](https://developers.data.dc.gov/marviewer/search/ssl/find?square=0542&lot=0089). 

ID=753 and 754 are a little different. These are also joined by a common record lot ([0711 0160](https://developers.data.dc.gov/marviewer/search/ssl/find?square=0711&lot=0160)), but two separate notices were issued for different buildings. We are deduplicating the notices in this case, however, because they have a common address. So we should not be double-counting the units when we look at the deduped records.

```
Prescat.Topa_notices_sales

        u_address_    u_dedup_
 ID       id_ref       notice        FULLADDRESS

 753      303694        Yes       130 M STREET NE   
 754      303694        No        130 M STREET NE   
```

eburton9:
@ptatian Thanks for this, I can write up a summary as you described but I'm looking at the code and realizing that I sumed the units for all notices and then I'm using that created var u_sum_units as the var in the proc tabulate (which includes duplicated notices) without using a where statement for tables 3a and 3b. I compared new output to old output and still in the difference of thousands between years for Ward 6 but not as drastic. I also added u_notice_with_sale=1 tables 6a and 6b. Apologies on the oversight on this!

ptatian:
Thanks for catching it! Yes, we should be using deduplicated notices for all except the first couple of tables.

eburton9:
@ptatian I'm not sure why my push isn't showing up on this pull request but I added a proc print to [an Excel for the three IDs](https://urbanorg.box.com/s/dmmfnfxcie1gsjob9jyakcosqfa25464) we were going to share with Martha and Ryan. I also started writing up a [summary ](https://urbanorg.box.com/s/eb2dtwm7menwvw1x6ehdmeuqlrw9y0r9) but I left some comments for you in the doc (the tables in the Excel are copied in the document as well. In general, the notices seem to have multiple SSLs in them rather than having a common record lot.

Prog/TOPA/TOPA_desc_tables.sas
```diff
@@ -42,6 +42,46 @@ proc sort data=TOPA_unit_check;
   by Ward2022 id;
 run;
 
+** Comparing SSLs from realprop dataset unit counts in addresses dataset by notice IDs **;
+data TOPA_entire_prop;
```
ptatian, line 46: This is a problematic merge because both data sets have multiple obs for each ID. The Topa_addresses data seems sufficient since it has the addresses and unit counts. If you want to show the property transactions, I think that is a separate Proc Print.

ptatian:
@eburton9 Thanks for sharing this. I made some edits/comments in the Word doc. Comment above about the merged data. Let me know if you have questions.

eburton9:
@ptatian I see, I updated the word document and proc print. I have a couple questions but I'll ask you at the check in.

eburton9:
@ptatian I added a proc print to output an excel sheet for Ward 6 real property # units from 2012 to 2014 (separated by tab). There were only two notices issued in 2013, which is likely why the units change so much across the years. Also I uploaded the updated tables to [the google drive](https://docs.google.com/document/d/1HI1GIsZ-ln9DbVgrZGQxjDuAP86n7vbC/edit?usp=sharing&ouid=107836493988618425982&rtpof=true&sd=true) but the formatting looks different than your previous version, not sure if that's a concern.

Issue #377, opened 2023-06-28
Fill in missing sales for TOPA notices


ptatian:
@eburton9 Based on some investigation of issues raised by the CBOs (see issue #375), we need to adjust our approach for matching sales to properties in Prescat.Topa_notices_sales. Here are a couple of challenges and my thoughts on corrections. **This issue should only be be started after finishing issue #376.**

### Sale date is missing or out of range

In some cases, we are missing a sale because the entry in the OTR data (Realprop.Sales_master/Prescat.Topa_realprop) is missing the date of sale. Examples of this are ID=316 and ID=232. Also, our sale cutoff date of May 2022 is omitting things where actual sales took place, such as ID=1380. 

To address this, for cases where the sale date is missing, I believe that we can use ownerpt_extractdat_first as the date instead. This is the date when that observation first appeared in the OTR data. Please make the following changes in the Sales_by_property_dates data step in TOPA_match_notices_property.sas.

1. Make a new sale date variable that = saledate if not( missing( saledate ) ), and = ownerpt_extractdat_first otherwise. 
2. Create an actual_saledate variable, that = 1 if we are using saledate and = 0 if we are using ownerpt_extractdat_first. 
3. Change the WHERE statement in that same datastep to an IF statement; use the new sale date variable you created in 1 instead of saledate; and change the upper bound from '31may2022'd to '31mar2023'd.
4. Label new variables. Assign the dyesno. format to actual_saledate. 

Use the new sale date variable you created for all the code that follows, including the Combo data step where the notice and sale data are combined. But please keep the original saledate and ownerpt_extractdat_first variables in the final Topa_notices_sales data set. 

To see what effect the above is having, please print out the sales in the revised data set for ID's referenced above. 

There are a few other things to look at, but let's start here and see where this gets us. Let me know if you have any questions.

eburton9:
\\sas1\dcdata\Libraries\PresCat\TOPA\TOPA_match_notices_propery.sas

Issue #375, opened 2023-06-24
Review CBO flagged TOPA properties


ptatian:
Several properties were flagged by CBOs reviewing the TOPA data as potential data problems. Need to investigate further.

_Properties listed as without sale that may have sold_

CNHED database unique notice ID | Unique property ID (DC MAR address ID) (Urban created var) | Notice offer of sale date (Urban created var) | All street addresses
-- | -- | -- | --
232 | 223601 | 09/23/2008 | 4000 Kansas Avenue NW
316 | 301911 | 01/28/2010 | 460 L Street NW
894 | 237132 | 04/15/2015 | 76 M Street NW

_Should be an earlier sale listed?_

CNHED database unique notice ID | Unique property ID (DC MAR address ID) (Urban created var) | Notice offer of sale date (Urban created var) | All street addresses
-- | -- | -- | --
986 | 225073 | 1/29/2016 | 930, 940, 960 Randolph Street NW
990 | 239117 | 2/5/2016 | 1111 Massachusetts Avenue NW

_Duplicate notice that should be removed?_
CNHED database unique notice ID | Unique property ID (DC MAR address ID) (Urban created var) | Notice offer of sale date (Urban created var) | All street addresses
-- | -- | -- | --
10004 | 156407 | 8/21/2015 | 4212 E. Capitol Street SE
10005 | 147608 | 8/26/2015 | 4212 East Capitol Street SE

PR #380, opened 2023-07-05
editing saledate vars

eburton9:
@ptatian I tried a couple different things to see why my code isn't working correctly but I'm unable to identify any issues so sending your way for help. Thank you.

### DIFF
--- a/Prog/TOPA/TOPA_match_notices_property.sas
+++ b/Prog/TOPA/TOPA_match_notices_property.sas
@@ -87,6 +87,10 @@ data Sales_by_property;
   if missing( ssl ) then delete;
 run;
 
+proc print data=Sales_by_property; /* can't find 894 ID */ 
+  where id=894;
+run;
+
 data Sales_by_property_dates;
   set Sales_by_property; 
   format all_saledate MMDDYY10.;
@@ -103,8 +107,13 @@ data Sales_by_property_dates;
   label actual_saledate='Property sale date used';
   if '01Jan2006'd <= all_saledate <= '31mar2023'd;
 run;
+%File_info( data=Sales_by_property_dates, printobs=5 ) 
 
-%File_info( data=Sales_by_property_dates, printobs=5 ) /** 4958 obs**/
+proc print data=Sales_by_property_dates;
+  where id in ( 232, 316, 894, 986, 990, 10004, 10005 );
+  by id;
+  var id ownerpt_extractdat_first SALEDATE all_saledate actual_saledate All_street_addresses u_address_id_ref;
+run;
 
 proc sort data=Sales_by_property_dates out=Sales_by_property_nodup nodupkey;
   by u_address_id_ref descending saledate;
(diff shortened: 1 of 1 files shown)


ptatian:
This syntax doesn't work with IF. 
`if all_saledate < '01Jan2006'd or > '31mar2023'd then delete;`

You need to use this:
`if all_saledate < '01Jan2006'd or all_saledate > '31mar2023'd then delete;`

Or, alternatively, you can use:
`if '01Jan2006'd <= all_saledate <= '31mar2023'd;`

IF without a THEN keeps only those observations that satisfy the condition.

ptatian:
Thanks, @eburton9 Does the revised code seem to be adding sales that we need for the notices I flagged in issue #375?

eburton9:
@ptatian It looks like for ID 232 and ID 316 there are dates from owernpt_extractdat_first when saledate was blank. ID 894 isn't coming up in this recently created dataset or the one before it.. I know the other ids are other issues but I printed them anyways. For 989 and 990 the ownerpt_extractdat_first is actually after the saledate. 10004 and 10005 do seem duplicated.

ptatian:
Thanks, @eburton9 Good that 232 and 316 now have sales showing up. 894 was a different problem. As was 989 and 990. And the duplicate. For these, we need to make changes to TOPA_data_to_DC_data.sas. I will make a new issue for that and tag you. We will need to complete the new issue and then return to this one.

eburton9:
@ptatian Updated based on [Issue 377](https://github.com/NeighborhoodInfoDC/PresCat/issues/377)

ptatian:
@eburton9 For your awareness, I've updated the Topa_addresses data set to remove non-exact adress matches. We were previously matching to some other address than the one listed, which is not what we want. This should result in these notices being dropped from the analysis.

   ID  | All street addresses       
-- | --
   25 | 4837 3rd Street NE         
   41 | 1302 12th Street NW        
  127 | 4209 Dix Street NW         
  176 | 4837 3rd Street NE         
  223 | 4912 Nash Street NW        
  309 | 1011 7th Street NW         
  471 | 2020 19th Street SE        
  617 | 1911 18th Street NW        
  812 | 501 12th Street SE         
 1016 | 4619 Nash Street NE        
 1381 | 4725 Foote Street NE       
10006 | 86 Webster Street NW       
10009 | 624 Girard Street Northwest

ptatian:
Hi @eburton9 I think that we may be done with this issue. If you would just check things run properly after the other data sets were updated, then I think we can finalize this one. We'll have to make some other changes based on the CBO review, but I'd like to finalize this first and go back to those. 

One small change is we should add a LABEL statement to the Proc Summary at line 33 to label the new u_sum_units variable. Currently, it's inheriting the label from Topa_addresses, but this label is a little misleading/confusing because it refers to individual buildings, where u_sum_units is the total of units across all buildings associated with the notice. I think a new label like, "Sum of MAR units for all associated buildings (Urban created var)" would be better. Thanks!

eburton9:
@ptatian updated label and there were zero obs when trying to print those IDs. I'm assuming that means the data update worked. I'll merge, sync, batch submit.

PR #382, opened 2023-07-06
manual fixes to ids 894, 10004, 10005

eburton9:
@ptatian Edits seem to work

### DIFF
--- a/Prog/TOPA/TOPA_data_to_DC_data.sas
+++ b/Prog/TOPA/TOPA_data_to_DC_data.sas
@@ -20,7 +20,7 @@
 %DCData_lib( MAR )
 %DCData_lib( RealProp )
 
-%let revisions = Add Ownercat and related vars to Topa_realprop.;
+%let revisions = Manually add SSLs and address_ids to notices;
 
 ** Download and read TOPA dataset into SAS dataset**;
 %let dsname="&_dcdata_r_path\PresCat\Raw\TOPA\TOPA-DOPA 5+_with_var_names_3_20_23_urban_update.csv";
--- a/Prog/TOPA/TOPA_data_to_DC_data.sas
+++ b/Prog/TOPA/TOPA_data_to_DC_data.sas
@@ -150,19 +150,40 @@ proc print data=TOPA_geocoded;
   where id in (10004, 10005);
 run;
 
-** Manually add SSL to ID 894 **;
+** proc print for manual adds **; 
+proc print data=Realprop.Parcel_base;
+  where lowcase( premiseadd ) contains "1350 fairmont";
+  id ssl;
+  var premiseadd;
+run;
+
+proc print data=Prescat.Topa_notices_sales; 
+  where id in (1159, 1543);
+run; 
+
+** Manually add SSL to IDs (in order) 894, 347&376, 387, 403&1075, 824&1383&1387&1579, 894, 1159&1543, 1466**;
 data Address_ssl_xref_new_obs;
   length ssl $ 17;
   infile datalines dlm=",";
   input address_id ssl;
 datalines;
-237132, 0620    0893
+237132 , 0620    0893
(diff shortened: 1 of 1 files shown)


ptatian:
@eburton9 Good! Now we need to add some additional parcels that were missing. This will take a bit of leg work. The notices currently don't have any parcel listed in Topa_ssl. 

 ID |  All_street_addresses
-- | --
 347 | 705 4th Street NW
 376 | 705 4th Street NW
 387 | 1615 Benning Road NE
 403 | 3341-3353 23rd Street, 2276 Savannah Street SE
 824 | 7428-7434 Georgia Avenue NW
 894 | 76 M Street NW
1075 | 3341, 3343, 3345, 3347, 3349, 3351, 3353, 3371, 3373 23rd Street SE & 2270, 2272, 2276 Savannah Street SE
1159 | 1350, 1352, 1354, 1356, 1358 Fairmont Street NW
1382 | 7428-7434 Georgia Avenue NW
1387 | 7428-7434 Georgia Avenue NW
1466 | 1208 M Street NW
1543 | 1350, 1352, 1354, 1356, 1358 Fairmont Street NW
1579 | 7428, 7430, 7432, 7434 Georgia Avenue NW

The trick is to find the matching parcels and create additional address_id, ssl rows to add to the datalines. 

Unfortunately, the MAR itself isn't any help here. What worked for me was searching through the Realprop.Parcel_base. You can do this with a Proc Print. For example, for 705 4th Street NW, I search for the first parts of the address. 

```
proc print data=Realprop.Parcel_base;
  where lowcase( premiseadd ) contains "705 4th";
  id ssl;
  var premiseadd;
run;
```

Which produces.

```
    SSL         PREMISEADD

0529    0037    705 4TH ST NW                     
0529    0848    0705 4TH ST NW                    
3281    0029    6705 4TH ST NW WASHINGTON DC 20012
3305    0020    4705 4TH ST NW WASHINGTON DC 20011
```

Two results that match the address we are looking for, 0529    0037 and 0529    0848. (Addresses in Parcel_base aren't formatted very consistently, so you don't want to over specify what you are looking for. That means you may get some false matches you need to leave out.) 

Then find the address_id that corresponds to this notice. If there are multiple addresses, use the u_address_id_ref value in Prescat.Topa_notices_sales. In this case, 242832. Then you can add this to datalines.

```
242832 , 0529    0037
242832 , 0529    0848
```

Do this for the rest of the notices in the list. Note that you only need to add the address_id and ssl combo once. If you have repeated addresses (like 347 and 376), you can skip the second notice. 

Let me know if you have questions. Thanks!

ptatian:
A couple ways. You could run the address through the [online MAR geocoder](https://developers.data.dc.gov/marviewer/home) and get the address_id that way. Or, you could look up the notice ID in Prescat.Topa_notices_sales.

ptatian:
The SSL won't be in Topa_geocoded because that is based entirely on MAR and MAR has no SSLs for these addresses. That's the problem we are trying to solve.

eburton9:
@ptatian Added the missing SSLs and address_ids. I created a [spreadsheet in our Box folder](https://urbanorg.box.com/s/t95qpfqu5blo834wvroijn7qqsirfmwa) with what goes with which ID before copying it into the datalines. I also double checked the TOPA_SSL dataset and they look to be there and correct!

ptatian (changes requested):
Hi @eburton9 This looks great! I appreciate the checking. Everything seems to be working fine. Please just update the %let revisions = message at the beginning to explain the changes we are making, and then you can go ahead and merge and batch submit on SAS1. You can then return to issue #377. 

I did find some geocoding issues and have a fix that I'll roll out after you merge this pull request. You will notice that we will have fewer notices with matching addresses after that. That's because currently we have some address matches that aren't exact and we need to filter those out.

Issue #374, opened 2023-06-13
Add affordable housing variables to TOPA database from Preservation Catalog


ptatian:
Hi @eburton9 Here's the next issue on the TOPA study data. We need to add variables for housing subsidies to the TOPA notice data. This corresponds to item 7.d in the data analysis plan. For each unduplicated notice, we want to define the following:

- Number of affordable units with LIHTC placed in service before unduplicated notice date
- Number of affordable units with LIHTC placed in service within one year after unduplicated notice date
- Number of affordable units with federal project-based subsidies (section 8, etc.), placed in service before unduplicated notice date
- Number of affordable units with federal project-based subsidies, placed in service within one year after unduplicated notice date
- Number of affordable units with DC HPTF subsidies awarded within one year after unduplicated notice date
- Date of formation of limited equity cooperative

These data can be obtained from the Preservation Catalog, specifically, PresCat.Subsidy. (See file metadata 
 //sas1/DCData/Metadata/meta_prescat_subsidy.html) This file contains separate subsidy records for Preservation Catalog projects. The subsidy type is identified by the PROGRAM and PORTFOLIO variables. For LIHTC, you can use `portfolio = "LIHTC"`; for federal-project based, you can use `portfolio in ( "202/811", "PB8", "PRAC" )`; for HPTF, you can use `portolio in ( "DC HPTF" )`; for limited equity coops, you can use `portfolio = "LECOOP"`.

The date when the subsidy began is POA_START_ORIG or, if it is missing, POA_START. 

The trick is merging this information with the TOPA notices. You should use Proc SQL for this. The first step is identifiying which Catalog projects are represented in the TOPA data. Catalog projects are uniquely identified by NLIHC_ID. PresCat.Building_geocode has address IDs for all the buildings associated with a NLIHC_ID. You want to connect the TOPA notices to the one or more NLIHC_ID values by matching the TOPA addresses with the Preservation Catalog addresses. So this is another many-to-many match (full join) that requires SQL. 

Once you have a crosswalk with TOPA notice IDs and NLIHC_ID, you can merge on the subsidy data. Again, you'll want to use SQL so that you can set match conditions of the subsidy start date either being before or up to one year after the notice. (You'll recall we did something like this when we matched the property sales with the notices.) 

Well, I know that's a bit much so think about it and we can talk further. Thanks!

eburton9:
[...]

PR #387, opened 2023-07-18
Match subsidy dataset to TOPA notices

eburton9:
@ptatian My first pass at Issue #374. In your instructions I believe you wanted the subsidy start date condition (before or up to one year after the notice) in the proc sql but I realized I needed to create the number of days between the subsidy and the notice after creating the new variable for the start of the subsidy. Doing a data step after the two proc sqls was what made sense to me but if that step can be incorporated into the proc sql let me know and I can work on that. Second, I noticed that for the same property there was date information for one of the subsidies but the same information was blank on the following row. For example, for ID 23 the section 8 dates are there but the DC HPTF dates are missing for the same property. Not sure if that's a concern or not but thought I would flag.


ptatian:
Thanks, @eburton9 This is a great start. But you don't want to begin by matching with Topa_notices_sales, which has one address ID per notice. You should match with Topa_addresses, which has the full list of addresses for each notice. We should get more matches that way. 

The matching of Subsidy to the notices seems to be fine. I would not keep matches if all_POA_start is missing, however. And you only need to keep the subsidy portolios that I identied in the issue. 

Yes, there may be ways to do everything in the Proc SQL, but I think using a Data step makes sense. 

I would add the Subsidy_id variable to the data you pull from Prescat.Subsidy. Because we may get multiple matches, we'll need a way to end up with unique unit counts. Subsidies are uniquely identified by Nlihc_id and Subsidy_id, so we can use that remove duplicate obseravtions in the end. 

Thanks! Let me know if you have any questions.

ptatian:
Yes, that makes sense. But to match TOPA notices to Nlihc_id's you need to use the full address lists in both data sources. Then you can merge in the sale date from Topa_notices_sales. So, I guess it is an additional match.

eburton9:
@ptatian Updated based on your comment. It does seem like there are a lot of Nlihc_id and subsidy_id duplicates.

ptatian:
@eburton9 I added a proc sort to remove the redundant matches. 

This is working well! Would you please create the variables that I listed in issue #374? I don't know that we want to keep the "within one year" condition after all, so please take that out for now.

eburton9:
@ptatian Sorry, are you talking about this list? 
- Number of affordable units with LIHTC placed in service before unduplicated notice date
- Number of affordable units with LIHTC placed in service within one year after unduplicated notice date
- Number of affordable units with federal project-based subsidies (section 8, etc.), placed in service before unduplicated notice date
- Number of affordable units with federal project-based subsidies, placed in service within one year after unduplicated notice date
- Number of affordable units with DC HPTF subsidies awarded within one year after unduplicated notice date
- Date of formation of limited equity cooperative

ptatian:
Yes

eburton9:
@ptatian It looks like there aren't any DC HPTF or federal subsidies within a year after the TOPA notice.. I also expanded the upper end of the limit to 1,000 and it doesn't seem like there's any. Is there something earlier in the program causing that, or is that realistic?

ptatian:
@eburton9 That does make sense. The HPTF portion should change once Donovan gets more of that data into the Catalog -- we have very little of it right now. But the project-based federal subsidies have been very limited in the last 40 years. 

How does it look for LIHTC? That is where we would expect to see some post-notice subsidies.

eburton9:
@ptatian Yep - looking at the file info, 10 properties had post notice LIHTC, which sums to 872 units.

ptatian:
Interesting. We'll have to compare that to what the CBOs flagged.

ptatian:
Is this ready for me to review?

eburton9:
@ptatian Yes, ready for review

ptatian (commented):
@eburton9 This is working well. You are following what I said in the variable list, but after reviewing the results I'd like to make a few changes. 

- **For notices with a date of sale**, let's use the date of sale for calculating u_days_notice_to_subsidy rather than the notice date. I think that is the more relevant date in those cases. 
- The count of units with subsidies after the sale/notice date **should include units that were in place before the date, too, provided those subsidies are still active**. In other words, the number of assisted units should carry over, as long as they are still assisted. To test whether the subsidy is still active, use the poa_end_actual var from Prescat.Subsidy. If poa_end_actual is not missing and is within one year after the sale/notice date, then don't count those units as "after" units. 
- I know that I said to add the date when the LE coops were formed, but really what we want is **the number of LE coop units before and after the sale/notice date**. So, similar to what we are doing to for the other subsidies, using the start date for the LE coop.
- I made some separate comments here on variable names.
- The final data set that we want to save should have **only one obs per notice (ID)**. So you want to aggregate the data by ID and take the sums of the before/after variables that you are creating. (This means we aren't retaining all the other variables that we used to create those unit counts.)
- Remember to label all the new variables! 

Thanks, Elizabeth. Let me know if you have any questions.

eburton9:
Thanks, @ptatian Quick question, are the N values considering missing? There's a lot of Ns in poa_end_actual but not(missing)poa_end_actual) doesn't seem to consider them missing (could be bad code on my end though) and I haven't found any code online to set values as missing.

ptatian:
Yes, N (actually .N) is a missing value. In SAS, . is the default missing but you can have .A, .B, .C, etc. I sometimes use .N as "n/a" or something like that. 

To test for a specific missing value, you need to write it out (e.g., x = .N), but if you want to test for any missing values you can use the missing( ) function (e.g., missing( x )).

eburton9:
@ptatian I'm having a bit of trouble. I tried to total the units with a data step using first/last, retain and if/else. It's only working for LECs because the other IDs have a combination of LIHTC and federal subsidies and first/last doesn't capture those. I was looking for a way to retain all but the last.id but couldn't find anything. So then I did a proc sql to just sum the units across the after and before vars then I could total it across portfolio type. Which is the better avenue? I still need to put in the poa_end_actual condition too but at first glance it seems like all the obs with the poa_end_actual are greater than a year past the sale date, so they would be included. Except, a lot of them had poa_end_actual as 2022 so now those would be expired?

ptatian:
Hi @eburton9 Proc Summary would be a better way to aggregate the data. It's simpler and more reliable. [Here's an intro](https://support.sas.com/resources/papers/proceedings-archive/SUGI90/Sugi-90-25%20Boyden.pdf).

eburton9:
@ptatian Ready for you to look at. I haven't filled out the finalize_dataset but I will once reviewed.

Prog/TOPA/TOPA_add_subsidies.sas
```diff
@@ -0,0 +1,169 @@
...
+  if portfolio = "LECOOP" and u_days_notice_to_subsidy < 0 then before_LEC_aff_units=Units_Assist;
+  if portfolio = "LECOOP" and 0 <= u_days_notice_to_subsidy <= 365 then after_LEC_aff_units=Units_Assist;
+  if missing(poa_end_actual) or u_sale_date-poa_end_actual > 365 or poa_end_actual > '01Jan2023'd; **not including units where subsidy's expired **;
+run;
+
+%File_info( data=TOPA_subsidy, printobs=10 )
+
+**Aggregating the data across rows with the same ID **;
+
+proc summary data=TOPA_subsidy
```
ptatian, line 113: When using a CLASS statement to specify the variables for aggregation, you usually want to add the NWAY option to the Proc Summary statement. By default, Proc Summary with CLASS creates intermediate totals as well as a grand total, which is not what we want. That's why the first obs in Topa_sum_rows is ID=.. That is the grand total of all rows. NWAY specifies that you only want the breakdowns using all CLASS variables at once, which is what we want here.

Prog/TOPA/TOPA_add_subsidies.sas
```diff
@@ -0,0 +1,169 @@
...
+ noprint;
+ var before_LIHTC_aff_units after_LIHTC_aff_units before_fed_aff_units after_fed_aff_units before_DC_HPTF_aff_units 
+	after_DC_HPTF_aff_units before_LEC_aff_units after_LEC_aff_units;
+ class id;
+ output out=TOPA_sum_rows
+	sum=sum_before_LIHTC_aff_units sum_after_LIHTC_aff_units sum_before_fed_aff_units sum_after_fed_aff_units sum_before_DC_HPTF_aff_units
```
ptatian, line 119: This is fine as is but, just FYI, you could have just said SUM= without the variable list and that would have used the same variable names in the VAR statement as the names of the new variables. Since we don't need the disaggregated versions, that would have been OK here.

Prog/TOPA/TOPA_add_subsidies.sas
```diff
@@ -0,0 +1,169 @@
...
+	sum_after_DC_HPTF_aff_units sum_before_LEC_aff_units sum_after_LEC_aff_units;
+run;
+
+%File_info( data=TOPA_sum_rows, printobs=10)
+
+**Summing the data across columns by subsidy type**;
+
+data TOPA_sum_total; 
+  set TOPA_sum_rows; 
+  if not(missing(sum_before_LIHTC_aff_units)) and not(missing(sum_after_LIHTC_aff_units)) then total_LIHTC_aff_units=sum_before_LIHTC_aff_units+sum_after_LIHTC_aff_units;
```
ptatian, line 129: I'm not sure why you are creating these totals. There's no reason to add the before and after numbers together and, in fact, that sum isn't meaningful because some units are going to be both before and after, so we'd be double counting.

Prog/TOPA/TOPA_add_subsidies.sas
```diff
@@ -0,0 +1,169 @@
...
+    else if missing(sum_after_DC_HPTF_aff_units) then total_DC_HPTF_aff_units=sum_before_DC_HPTF_aff_units;
+	else total_DC_HPTF_aff_units=sum_after_DC_HPTF_aff_units; 
+ label total_DC_HPTF_aff_units='Total subsidy assisted units from the DC Housing Production Trust Fund (Urban created var)';
+
+  if not(missing(sum_before_LEC_aff_units)) and not(missing(sum_after_LEC_aff_units)) then total_LEC_aff_units=sum_before_LEC_aff_units+sum_after_LEC_aff_units;
+    else if missing(sum_after_LEC_aff_units) then total_LEC_aff_units=sum_before_LEC_aff_units;
+	else total_LEC_aff_units=sum_after_LEC_aff_units;
+ label total_LEC_aff_units='Total affordable units formed from Limited-Equity Cooperatives (Urban created var)';
+
+  keep id total_LIHTC_aff_units total_fed_aff_units total_DC_HPTF_aff_units total_LEC_aff_units;
```
ptatian, line 149: We don't want these total_ variables. We want the before_ and after_ variables.

ptatian (changes requested):
Thanks, @eburton9 You got the Proc Summary working. A few comments.

eburton9:
@ptatian Whoops, I had read some of the instructions wrong. I took out that last data step and made edits to the proc summary.

eburton9:
@ptatian I added a new data step for the after vars, which I think are hopefully working how they should. I checked 5-10 but may need to look again with a new set of eyes since there's a lot of different options. I'm not sure why my push isn't showing up in the pull request also..

eburton9:
@ptatian Ready for you to look at.

ptatian:
@eburton9 The way you had this part previously:

	else if not(missing(poa_end_actual)) and poa_end_actual-u_sale_date <= 365 then after_LIHTC_aff_units=.;
	else if not(missing(poa_end_actual)) and missing(u_sale_date) then after_LIHTC_aff_units=before_LIHTC_aff_units;

It would never get to the second condition, because if u_sale_date is missing, then poa_end_actual-u_sale_date would be missing (arithmetic with missing values returns a missing value), and missing is <= 365 (missing is represented in SAS as a very large negative number).

I rewrote replacing those two statements as:

	else if not(missing(poa_end_actual)) and poa_end_actual-max(u_sale_date,u_notice_date) <= 1825 then after_LIHTC_aff_units=.;

I use max() to take the later of u_sale_date and u_notice_date. Normally, u_sale_date is later, but if u_sale_date is missing, then it will use u_notice_date, which is never missing. 

If you can make that same change for the other subsidies, that would be great.

ptatian:
I use max() with u_days_notice_to_subsidy calculation, too, so that we are using the u_notice_date if u_sale_date is missing,.

eburton9:
@ptatian Thank you. That makes sense. I updated the other three subsidies.

ptatian (approved):

PR #415, opened 2023-08-21
Create adding new subsidy prog

eburton9:
@ptatian I think there is an issue with not having any zip codes.. I manually found five and re ran the prog and it worked for those.

### DIFF
--- a/Macros/Add_new_projects.sas
+++ b/Macros/Add_new_projects.sas
@@ -28,6 +28,7 @@
   input_file_pre=, /** First part of input file names **/
   input_path=&_dcdata_r_path\PresCat\Raw\AddNew,  /** Location of input files **/
   use_zipcode=Y,  /** Use ZIP code when geocoding (set =N when invoking macro if input data does not have ZIP codes) **/
+  matching_project_list=, /** Destination XLS file for list of projects matching existing Catalog or another new project **/
   address_data_edits=, /** Address data manual edits **/
   parcel_data_edits= /** Parcel data manual edits **/
   );
@@ -41,6 +42,7 @@
     input_file_pre=&input_file_pre,
     input_path=&input_path,
     use_zipcode=&use_zipcode,
+    matching_project_list=&matching_project_list, 
     address_data_edits=&address_data_edits,
     parcel_data_edits=&parcel_data_edits
   )
--- a/Macros/Add_new_projects.sas
+++ b/Macros/Add_new_projects.sas
@@ -27,6 +27,7 @@
 %macro Add_new_projects( 
   input_file_pre=, /** First part of input file names **/
   input_path=&_dcdata_r_path\PresCat\Raw\AddNew,  /** Location of input files **/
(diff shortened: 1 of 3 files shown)


ptatian:
@eburton9 The input data set had some embedded new line characters. I manually deleted those. 

Yes, the other problem was the missing ZIP codes. The SAS geocoder doesn't give good results if you ask it to geocode with ZIP codes (which was the default for this macro) but no ZIP codes are provided in the data. I've added an option to the macro, use_zipcode=, which can be set to N to geocode without ZIP codes. This gives better results. 

There are still 23 addresses that could not be matched. 4 have blank addresses. 1 is a scattered site project with two addresses (1913 Gallaudet St NE; 1625 Butler St SE), which should either be assigned to one address only or split into two projects. Several others have missing or incorrect quadrants or incorrect street types (e.g., "Ontario Court" instead of "Ontario Road"). These should be fixed manually.

ptatian:
But another issue is it seems many of these projects are not new but already in the Catalog. I'm looking at something to help with that.

eburton9:
@ptatian I updated the new_projects_issue_303 csv

There's 5 address issues left: 
ID 45: 2300 30th Pl NE https://octo.quickbase.com/db/bit4krbdh?a=dr&rid=41 new construction and I don't think the address is in MAR yet. 
ID 151: The address 4274 East Capitol St NE is the correct one for East Capitol Gardens but it's not working.
ID 384: I can't find an address for "Wagner Senior"  2422 Wagner St SE and Victory Square (600 Barnes St NE) came up when I googled it. 
ID 388: The address 2850 Douglass Pl SE is the correct one for City View Apartments but it's not working. 
ID 402: The address 2327 Good Hope Rd SE is the correct one for Woodmont Crossing but it's not working. 

Also flagging an old id that I took out: 14 Florida Ave doesn't exist yet. It's an empty lot at 10 Q St NW, which is the address of the ID directly above it so I removed it.

ptatian:
Thanks, @eburton9. I'll look at the addresses that aren't working. The geocoder has some quirks. 
I've pushed some changes to look more closely at the projects that may already be in the Catalog. It appears many of them are. The program now produces a spreadsheet with that list (Prog\AddNew\New_projects_issue_303_matching_project_list.xls). I get a weird error message when I try to open the file in Excel, but if you just click OK the file will open.

ptatian:
Would you also be able to take a look at the `\\sas1\DCData\Libraries\PresCat\Raw\AddNew\New_projects_issue_303_subsidy.csv` file? That's not reading in properly, either, and I think it needs some editing.

ptatian:
@eburton9 The subsidy file columns should match this one: `\\sas1\DCData\Libraries\PresCat\Prog\AddNew\Templates\New_projects_issue_nnn_subsidy.xlsx`  Donovan might have been working off an older version.

ptatian:
@eburton9 Subsidy_Info_Source should all be "DC/AFFPIPELINE" and Subsidy_Info_Source_Date should be 7/1/2022.

eburton9:
@ptatian Okay, it seems it's running correctly now.

Issue #419, opened 2023-08-25
Create maps for TOPA report


eburton9:
1) Co-op map & table (Table 11 without condos) d_le_coop=Yes 2) Affordability added or preserved d_affordable=1. Note that this will change once we add the HPTF data. 3) tenant assoccations assign rights or become LEC, d_ta_assign_rights=1 or d_le_coop=1.

PR #424, opened 2023-08-25
adding two tables to correspond to maps

eburton9:
related to issue #419 @ptatian Added the two tables (28/29) that Martha wanted.


eburton9:
@ptatian I removed the u_notice_date =. 
So now the notice dates that Farah wanted taken out will show up in the TOPA_table_data spreadsheet that she's been using to check and in the desc tables. Should we communicate the reasoning to her? Unless I'm not following what you wanted to do correctly.

PR #474, opened 2024-12-06
Create 3 PresCat Update Programs for new HUD data

RAGarMir:


ptatian (approved):
@RAGarMir These look good! I pushed a couple of other changes to the output that is generated by a couple of programs. You can ignore changes to the other programs and just batch submit the 3 update programs that you created. Let me know if you have any questions. Thanks!

Issue #473, opened 2024-11-26
DC Preservation Study tabulations


ptatian:
From Urban SOW: 

Task as stated in CNHED grant agreement:
_A brief statistical summary of the current stock of affordable housing in the District of Columbia
derived from the HousingInsights.org and DC Preservation Catalog._

Urban will be responsible for preparing summary tabulations from the DC Preservation Catalog (which is
the source of information for HousingInsight.org). The Catalog includes data on assisted affordable
housing in the District, including federally-subsidized housing (public housing, project-based rental
assistance, HUD-insured mortgages, and low income housing tax credit properties) and locally-
subsidized housing (housing production trust fund, preservation fund). Urban will tabulate the number
of project, total units, and assisted units according to these characteristics:

- Location (ward, ANC, neighborhood cluster)
- Subsidy type
- Age of building
- Type of owner
- Neighborhood characteristics (demographics, housing market conditions and trends

ptatian:
@RAGarMir Thanks for checking in earlier this week. I wanted to provide a bit more guidance on this task. It may be helpful to refer to the [PresCat data structure diagram](https://urbanorg.box.com/s/1edpk8o9mvih4kua7vb02p227fkyurwc) while doing this work. 

I have to correct a couple of things I said when we met. First, it is better to use PresCat.Project_category_view rather than PresCat.Project for the project-level data. PresCat.Project_category_view already links to the geography data (Project_geocode), so you don't have to do that separately. 

Second, because projects can have multiple subsidies, and each subsidy can have its own count of assisted units, this can lead to uncertainty when you have overlapping subsidies as to how many total assisted units there actually are in a project. 

To address this, I have a program that combines the subsidies and project info: **Prog\Project_assisted_units.sas**. We should use this code as the basis for our tabulations. **Please make a copy of this program, and use it as the starting point for the Preservation Study tabulations.** **(Make sure you merge all the changes from main into your branch before copying the file as I recently pushed some changes to this program.)**

This code includes three proc tabulates. The 3rd one counts up projects and assisted units by ward, which is the first part of one of the tabulations that we need. You should copy this code to create similar tables for ANC (anc2012) and neighborhood cluster (cluster2017) instead of ward. This takes care of the first two tablulations in our list. 

 - Projects and assisted units by location (ward, ANC, neighborhood cluster)
 - Projects and assisted units by subsidy type

_Projects and assisted units by age of building_

[...]

PR #480, opened 2025-01-08
Create PresStudyTabulations.sas

RAGarMir:
This is the progress I have had so far on creating the tabulations for the first 3 tables. I have the shell of the code done and I am working through the errors. Currently working on errors when merging datasets Project_AOB & Realprop.Cama_parcel_2023_05 by SSL.

### DIFF
--- a/Prog/PresStudyTabulations.sas
+++ b/Prog/PresStudyTabulations.sas
@@ -320,7 +320,7 @@ data Project_Age_Of_Building;
  	 if inProject and inCama;
    
 run;
-
+***Getting the Message that this statement is out of order***;
 proc summary data=Project_Age_Of_Building; 
   by nlihc_id;
   output out=Project_Building_Age;
--- a/Prog/PresStudyTabulations.sas
+++ b/Prog/PresStudyTabulations.sas
@@ -142,7 +142,7 @@ data Project_assisted_units;
   set Project_subsidy;
   by NLIHC_ID;
   
-  retain num_progs total_units min_asst_units max_asst_units asst_units1-asst_units&MAXPROGS
+  retain num_progs min_asst_units max_asst_units asst_units1-asst_units&MAXPROGS
          poa_start_min poa_end_min poa_end_max compl_end_min compl_end_max proj_ayb_min
          min_asst_units_pb max_asst_units_pb poa_start_min_pb poa_end_min_pb poa_end_max_pb compl_end_min_pb compl_end_max_pb 
          min_asst_units_tc max_asst_units_tc poa_start_min_tc poa_end_min_tc poa_end_max_tc compl_end_min_tc compl_end_max_tc 
@@ -152,7 +152,6 @@ data Project_assisted_units;
   
   if first.NLIHC_ID then do;
   
-    total_units = .;
     num_progs = 0;
     
     min_asst_units = .;
@@ -198,8 +197,6 @@ data Project_assisted_units;
   
   num_progs + 1;
   
-  total_units = max( total_units, Proj_Units_Tot, Units_Assist );
-
   select ( portfolio );
     when ( 'PUBHSNG' ) a_aunits{&PUBHSNG} = sum( Units_Assist, a_aunits{&PUBHSNG} );
     when ( 'PB8' ) a_aunits{&S8PROG} = sum( Units_Assist, a_aunits{&S8PROG} );
@@ -262,12 +259,12 @@ data Project_assisted_units;
   if last.NLIHC_ID then do;
   
     do i = 1 to &MAXPROGS;
-      a_aunits{i} = min( a_aunits{i}, total_units );
+      a_aunits{i} = min( a_aunits{i}, Proj_units_tot );
     end;
 
-    max_asst_units = min( sum( of asst_units1-asst_units&MAXPROGS ), total_units );
-    max_asst_units_pb = min( max_asst_units_pb, total_units );
-    max_asst_units_tc = min( max_asst_units_tc, total_units );
+    max_asst_units = min( sum( of asst_units1-asst_units&MAXPROGS ), Proj_units_tot );
+    max_asst_units_pb = min( max_asst_units_pb, max_asst_units, Proj_units_tot );
+    max_asst_units_tc = min( max_asst_units_tc, max_asst_units, Proj_units_tot );
     
     mid_asst_units = min( round( mean( min_asst_units, max_asst_units ), 1 ), max_asst_units );
     mid_asst_units_pb = min( round( mean( min_asst_units_pb, max_asst_units_pb ), 1 ), max_asst_units_pb );
@@ -314,7 +311,7 @@ data Project_assisted_units;
   
   end;
   
-  format poa_start_min poa_end_min poa_end_max compl_end_min compl_end_max mmddyy10.;
+  format poa_start_min: poa_end_min: poa_end_max: compl_end_min: compl_end_max: mmddyy10.;
   
   drop i portfolio Units_Assist poa_start poa_end compl_end proj_ayb _freq_ _type_;
 
@@ -332,6 +329,8 @@ options missing='0';
 options nodate nonumber;
 
 %fdate()
+
+ods listing close;
 
 ods rtf file="&_dcdata_default_path\PresCat\Prog\PresStudyTabulations.rtf" style=Styles.Rtf_arial_9pt;
 
@@ -421,9 +420,9 @@ data Project_AOB;
     Project_assisted_units
 		(in=inProject)
     Prescat.Parcel
-		(in=inParcel);
+		(in=inParcel keep=nlihc_id ssl);
   by NLIHC_ID;
-  	if inProject and inParcel;
+  	if inProject;
    
 run;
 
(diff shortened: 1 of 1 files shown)


Prog/PresStudyTabulations.sas
```diff
@@ -0,0 +1,369 @@
...
+  merge
+   Project_AOB
+   	(in=inProject)
+   Realprop.Cama_parcel_2023_05
+   	(in=inCama);
+
+  by SSL;
+ 	 if inProject and inCama;
+   
+run;
```
ptatian, line 318: `ERROR: BY variables are not properly sorted on data set WORK.PROJECT_AOB.`

When doing a MERGE, all data sets must be sorted by the BY variables. Use [Proc Sort](https://documentation.sas.com/doc/en/proc/1.0/p0ha9ymifyaqldn14m2xlqw252wa.htm) to sort the Project_AOB data set by SSL before the data step that does the merge.

Prog/PresStudyTabulations.sas
```diff
@@ -0,0 +1,369 @@
...
+
+proc summary data=Project_Age_Of_Building 
+  by nlihc_id;
```
ptatian, line 321: `ERROR 22-322: Syntax error, expecting one of the following: ;, (, ...`

Missing a semicolon at the end of the Proc Summary statement.

Prog/PresStudyTabulations.sas
```diff
@@ -0,0 +1,369 @@
...
+    ;
+  table 
+    /** Rows **/
+    ( all='DC Total' AYB=' ' )
+    ,
+    /** Columns **/
+    sum='Assisted Units' * ( all='\b Total' ProgCat=' ' ) * mid_asst_units=' '
+    ;
+  format ProgCat ProgCat.;
+  format value year_built (notsorted)
```
ptatian, line 347: `ERROR 22-322: Syntax error, expecting one of the following: a name, a format name, ...`

The code to create a new user format needs to be a Proc Format (not a Format statement) which is separate from and comes before the Proc Tabulate code. Please refer back to the code I included in the issue. You then assign the year_built format you created to the AYB variable using a Format statement as part of the Proc Tabulate, such as is being done with `format ProgCat ProgCat.;`.

[You can look here](https://www.statology.org/sas-proc-format/) for other examples of using Proc Format.

ptatian (commented):
@RAGarMir Please see my comments on the errors you are getting. Let me know if you are still having challenges.

ptatian:
@RAGarMir Regarding this error.
```
399        ***Getting the Message that this statement is out of order***;
400        proc summary data=Project_Age_Of_Building;
401          by nlihc_id;
402          output out=Project_Building_Age;
403            min(AYB)=;
               ___
The SAS System

               180
ERROR 180-322: Statement is not valid or it is used out of proper order.
404        run;
```
The problem is that `min(AYB)=` is part of the output statement. Therefore, you should not have a semicolon after `Project_Building_Age`.

RAGarMir:
Hello @ptatian I removed that semicolon. I still kept getting an error at the Proc Summary statement saying the data set was not sorted properly. I changed "by nlihc_id" to "by SSL" and that seemed to get rid of the error, but, unless I was reading your instructions wrong, those said to sort by NLIHC_ID at this stage. Is there something else I may be missing here?

ptatian:
Hi @RAGarMir. Yes, sorry, I overlooked that. The input data set does need to be sorted by the BY variables. But we want to summarize by project (nlihc_id) and not by parcel, so changing to BY SSL is not what we want. 

There are two possible solutions. One is to use [Proc Sort](https://documentation.sas.com/doc/en/pgmsascdc/v_059/proc/n0sv3pcvpllvh9n1521qlriu6u4n.htm) to sort the data set Project_Age_Of_Building before the Proc Summary step. 

A second option is to change Proc Summary step as follows:
```
proc summary data=Project_Age_Of_Building nway;
  class nlihc_id;
```
Changing BY to CLASS removes the need to have the input data sorted. But CLASS by itself produces multiple summary levels. Adding the NWAY option to Proc Summary restricts to just the most basic summary. 

If the only reason we need Project_Age_Of_Building sorted is for the Proc Summary, then I would probably go with the second option.

ptatian:
Just to clarify, the rest of the Proc Summary step would be the same as before. Only the two lines I highlighted above need to be changed.

RAGarMir:
@ptatian I created and ran the code to create the table by owner type. The code tan fine without any errors, but the desired table is not part of the output, I am not sure how to debug in this situation.

ptatian:
@RAGarMir The proc tabulate needs to be placed before the ods rtf close statement. Only output from code between the first ods rtf… statement and the ods rtf close statement will be directed to the output file.

RAGarMir:
Hello @ptatian I currently have the ods rtf close statement currently comes after the proc tabulate statement but I am still not getting the output of the table in the word document.

ptatian:
Hi @RAGarMir Yes. You do have the Proc Tabulate in the right place now, but the problem is the Proc Tabulate isn't producing any output. If you look at the LOG, you will see these messages after the last Proc Tabulate.

```
NOTE: No observations were selected from data set WORK.PROJECT_ASSISTED_UNITS.
NOTE: There were 0 observations read from the data set WORK.PROJECT_ASSISTED_UNITS.
      WHERE (ProgCat not = .) and (not MISSING(parcel_owner_type));
```

Since no data were selected, no table was produced. 

So, the next step would be to go back to where PROJECT_ASSISTED_UNITS was created and see what might be going on. 

If we add parcel_owner_type to the list of freqvars= in the %File_info() macro, we get this in the output. 

```
parcel_
owner_                              Cumulative    Cumulative
type       Frequency     Percent     Frequency      Percent
------------------------------------------------------------
                541      100.00           541       100.00  
```

This tells us that all the values of parcel_owner_type are blank. Blank equates to "missing" for character variables in SAS and, therefore, the condition `not( missing( parcel_owner_type ) )` in the Proc Tabulate where statement is never true. 

So, the next step is to figure out what is going wrong with parcel_owner_type. If we look in the LOG after the Project_Owner_Summary data step, we see messages like this:

```
NOTE: Variable parcel_owner_type_111 is uninitialized.
NOTE: Variable parcel_owner_type_045 is uninitialized.
NOTE: Variable parcel_owner_type_115 is uninitialized.
etc.
```

This means that the variables parcel_owner_type_111, parcel_owner_type_045, etc. were never set to any value before they were used. We need to continue to trace back to see where the problem emerged. The step before this one was the Proc Summary, so let's look at that. 

```
***Summarizing new variables by Nlihc_id***;
proc summary data=parcel_owner_type nway; 
  class nlihc_id;
  output out=parcel_owner_type;
run;
```

As noted in the comment, the purpose of Proc Summary was to summarize the new variables, but you didn't include the new variables in the Proc Summary. You need to provide a VAR statement to do this. Also, you didn't provide any statistic keyword with the output statement, which means SAS will use SUM= as the default. But, we actually want MAX= here so that the values of the summarized variables will be equal to either 1 or missing. 

```
proc summary data=parcel_owner_type nway; 
  class nlihc_id;
  var parcel_owner_type_111 Variable parcel_owner_type_045 parcel_owner_type_115 etc.;
  output out=parcel_owner_type max=;
run;
```

Once that is fixed, the values of parcel_owner_type should be assigned properly. You should **check the data**, however, to make sure that the code is working the way we want. That is, you should look not only at the final output, but at some of the intermediate data sets so that you can be sure that things are working properly. You can find information on this in the [Examining Intermediate Results section](https://github.com/NeighborhoodInfoDC/SAS/blob/master/Guide/Debugging%20programs.mkdn#Intermediate) of the debugging guide.

ptatian:
I went through a lengthy description above so that you could see my process for tracking down errors. You need to **work your way backward** through the code to identify where the problem is occuring. Look at **intermediate results** (data sets, output) to see that everything is working as expected. Try to trace the problem back to its earliest point. That is the first part of finding a solution.

I linked to one part of it above but you should take a look again at the [SAS Debugging Guide](https://github.com/NeighborhoodInfoDC/SAS/blob/master/Guide/Debugging%20programs.mkdn). And [this tutorial](https://libguides.library.kent.edu/SAS/ViewData) explains how to view data sets either in the SAS Windows Viewtable window with Proc Print.

RAGarMir:
@ptatian I made the requested changes, additions, and fixes.

RAGarMir:
@ptatian the latest commit should have the code I added for the tables on neighborhood characteristics, earliest expiration dates, and size of development

RAGarMir:
@ptatian I consolidated projects and assisted units into one table for Sec8 and Lihtc tables

ptatian:
Hi @RAGarMir. Thank you for all your work on this. I made a few adjustments to the expiration tables, and also added code for outputting lists of expiring Sec 8 and LIHTC projects. In making these changes, I noticed that the year built/renovated tables were not using the project-level data, but a data set that had multiple records for each project. I fixed that by merging the summary AYB and EYB variables back onto the Project_assisted_units data. 

I'll approve all the changes for the review. I'd like you to finish this up, please, by merging the branch with main in the pull request, running the Sync.ps1 file in Windows Powershell to update the code on SAS1, and then doing one last batch submit of the program on SAS1. When finished, please copy the full pathname of the LOG file on SAS1 file to issue #473 and close the issue. 

Let me know if you have any questions. Thank you!

ptatian (approved):
Changes approved. Ready to be finalized.

Issue #485, opened 2025-07-09
Preservation Catalog PresCat July 2025 update


ptatian:
_Note: If you have not previously done so, [clone the PresCat repo](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/adding-and-cloning-repositories/cloning-and-forking-repositories-from-github-desktop#cloning-a-repository) to **C:\DCData\Libraries**. Pull the latest changes from main branch before proceeding._

## Create and test update programs

- [x] [Create new branch](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/making-changes-in-a-branch/managing-branches#creating-a-branch) **update-#**_n_ in GitHub

- [x] [Create and test Sec8MF update program](https://urbanorg.app.box.com/notes/45544495689)

- [x] [Create and test MFIS update program](https://urbanorg.app.box.com/notes/48593611845)

- [x] [Create and test REAC update program](https://urbanorg.app.box.com/notes/317426216726)

- [x] Review LOG and LST files for any problems.

- [x] [Request review and finalize programs](https://urbanorg.app.box.com/notes/295752223069)

- [x] After code has been reviewed and approved, merge branch, sync to `\\sas1\DCData\`.

## Final batch submits on SAS1

- [x] Batch submit `\\sas1\DCDATA\Libraries\PresCat\Prog\Updates\Update_Sec8mf_????_??.sas` (latest version created above)

- [x] Batch submit `\\sas1\DCDATA\Libraries\PresCat\Prog\Updates\Update_MFIS_????_??.sas` (latest version created above)

- [x] Batch submit `\\sas1\DCDATA\Libraries\PresCat\Prog\Updates\Update_REAC_????_??.sas` (latest version created above)

- [x] Batch submit ~~`\\sas1\DCDATA\Libraries\PresCat\Prog\Updates\Update_parcel.sas`~~ `F:\DCDATA\Libraries\PresCat\Prog\Updates\Update_parcel_remote.sas`

- [x] Batch submit ~~`\\sas1\DCData\Libraries\PresCat\Prog\Updates\Update_real_property.sas`~~ `F:\DCData\Libraries\PresCat\Prog\Updates\Update_real_property_remote.sas`

- [x] Review LOG and LST files for any problems.

[...]

PR #501, opened 2025-08-04
Preservation Catalog PresCat July 2025 update #485

RAGarMir:
@ptatian I created the update files and edited them for the remote process, this just involved changing the include system. The Reac and Sec8MF ran without errors and they created the necessary files. However the MFIS file ran with multiple errors. The program could not find various variables. I did not make any changes to Macros.

### DIFF
--- a/Macros/Update_LIHTC.sas
+++ b/Macros/Update_LIHTC.sas
@@ -17,7 +17,7 @@
 
 %macro Update_LIHTC( 
   Update_file=, 
-  Finalize=Y,
+  Finalize=N,
   Project_except=Project_except,
   Subsidy_except=Subsidy_except,
   Manual_subsidy_match=,
--- a/Macros/Update_LIHTC_finish.sas
+++ b/Macros/Update_LIHTC_finish.sas
@@ -35,6 +35,7 @@
   
   %Finalize_data_set( 
     /** Finalize data set parameters **/
+    finalize=&finalize,
     data=Subsidy,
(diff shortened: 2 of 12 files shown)


ptatian:
Thank you, @RAGarMir. Yes, the problem OK, the problem with the MFIS data is because of the property_street variable that HUD dropped from the files. 

In the **Macros\Update_MFIS_init.sas** file, there are two places where property_street is referenced. The first is

```
  %let Subsidy_missing_info_vars = 
      &Subsidy_Info_Source_ID_src Premise_id Property_name Property_street SOA_cat_sub_cat
      ;
```

Like you did in the HUD macro Macros\MFIS_read_update_file.sas, we will use %if-%then-%else logic to create different versions of this variable list. 

```
	%if %sysevalf( &filedate ) >= %sysevalf( '31jul2016'd ) %then %do;
          %let Subsidy_missing_info_vars = 
            &Subsidy_Info_Source_ID_src Premise_id Property_name SOA_cat_sub_cat
          ;
      %end;
      %else %do;
          %let Subsidy_missing_info_vars = 
            &Subsidy_Info_Source_ID_src Premise_id Property_name Property_street SOA_cat_sub_cat
          ;
      %end;
```

The first one, which applies to files from this point onward, does not have property_street, while the list for earlier files does. 

Please do the same thing for the second place where Property_street appears in the macro program. 

That should fix the error. Please let me know if you have any questions. Thank you!

